PR #27, opened 2018-11-16
DO NOT MERGE WIP: Joyent case study

ZooeyMiller:
Made a WIP PR so that the preview is viewable, no mobile or tablet designs so this is blocked until I've got those.


ZooeyMiller:
in light of all of the newness i am deleting this pr and we can use it as a basis for the joyent page if needed with the new grid.

PR #47, opened 2018-11-22
Sets up the environment for storyshot tests

ZooeyMiller:
Making a separate PR so I can pivot to work on a bug and continued test writing can be done on another PR.


SaraVieira:
Can you add this tests to circle city?

PR #50, opened 2018-11-22
Update type scale

SaraVieira:
https://trello.com/c/GcZ4qmku/127-update-typescale


SaraVieira:
@JhonnyYLD I think it was because H1 and H2 had the same design

PR #71, opened 2019-01-02
Create our work page

jsms90:
## Description - [Trello ticket](https://trello.com/c/XCR95mnl/19-our-work-case-studies)

- `our-work.js` pulls all case studies from contentful & renders page title + loops through case studies and handles layout  
Note: currently adding speciality (either 'Design' or 'Engineering') to each case study within codebase - should be added to contentful.
- Created `GreyBackgroundWithoutOffset` (unsure what's involved in [this trello ticket](https://trello.com/c/P5FkCAfD/193-refactor-gray-background-component-to-be-simplaer-and-not-account-for-offsets-we-no-longer-need)
- `/OurWork/CaseStudy.js` handles layout of a particular case study, which is different from the `caseStudy` component used on the homepage & `topCaseStudy` in "Common" - might be worth amalgamating these in future, if we can?
- Removed `<hr>` styling for globalStyle & work-stage/elements.js. Replaced with a new `Hr` component within the Common folder.
- Made background colour of navbar within `layout.js` more customisable, to allow the colour to be grey (previously either blue or white)

## Review template

The below is for reviewers of this PR to copy/paste into the review body and fulfill.

- Old grid break points: 0-599, 600-1095, 1095+
- New grid break points: 0-470, 471-552, 553-700, 701-899, 900-1196, 1197+

I have opened the preview link and:

- [ ] checked that this PR resolves the spec provided from trello / this description - What is meant by "spec"? Copy does not quite match what's in the google doc yet, as I wanted to check with someone before changing the content within Contentful
- [x] checked the effected pages at all breakpoints
- [x] ensured that this PR does not introduce any obvious bugs - Checked the pages that currently use an `<hr>` (for example), since I know I'm altering that. But what is usually done at this step?

PR #80, opened 2019-01-09
feat/add join us

antoniocapelo:
## Review template


src/components/GreyBG.js
```diff
@@ -27,58 +27,60 @@ export default styled.section`
...
-    margin-top: ${({ topOffset }) =>
-      remcalc((topOffset && topOffset * -1) || PADDING)};
-    margin-bottom: ${({ topOffset }) => remcalc(topOffset || -PADDING)};
-    &:after {
-      ${afterStyles}
-      height: ${({ topOffset }) =>
-        remcalc(topOffset ? topOffset * -1 : PADDING)};
-    }
-  `}
+  ${isNot('static')`
```
antoniocapelo: `static` is meant to be a prop to disable the negative margins that currenlty the `GreyBG` applies by default. I visited some pages in the website and see why this is needed. So I thought a single prop could be a good solution to disable that behavior when needed.

src/components/JoinUs/Learning.js
```diff
@@ -0,0 +1,77 @@
...
+  <Section greyBg>
+    <Row>
+      <FirstColumn>
+        <H2>{title}</H2>
+      </FirstColumn>
+      <SecondColum width={[1, 1, 1, 1, 6 / 12, 5 / 12]}>
+        <List>
+          <ReactMarkdown
+            renderers={{
+              heading: props => <Paragraph noMargin bold {...props} />, // eslint-disable-line react/display-name
```
antoniocapelo: solution for content creators to reacreate the following layout with just a heading and a paragraph:

image

antoniocapelo:
Some notes regarding this MR:

1. Created compensated cols to better adapt to current design specs (gutter width)
Our design specs show our gutter width at around 48px but our current `Col` components use a 48px each, which causes our grid is visually the double of what it needs (48+48). 

I don't know if this is a known issue or not so until we can establish a convention, I created the `CompensatedCol` and `CompensatedRow` (probably not the best names, but I'm hoping they're temporary) that allow for the single gutter width. 

2. Filter for the Open Position section are not implemented yet since there's still some discussion going around this section details and data coming from Lever

3. In https://github.com/yldio/yld.io/pull/71 @jsms90 created a separate component for the GreyBg without negative margins — I used a different approach where a prop change its behavior. I can accommodate to Jen's solution, if needed.

4. I'm not displaying the link to 'our career framework' since there's no page yet

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,97 @@
...
+  >
+    {children}
+  </ShiftedColumn>
+)
+
+export const List = ({ list }) => (
+  <Padding top={1}>
+    <ReactMarkdown
+      renderers={{
+        heading: props => <Paragraph noMargin bold {...props} />, // eslint-disable-line react/display-name
```
jsms90: I reckon this should use a heading component (not sure which size), rather than `Paragraph`?

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,97 @@
...
+import GreyBG from '../GreyBG'
+
+/**
+ * Layout
+ */
+export const FirstColumn = ({ children }) => (
+  <Col width={[1, 1, 1, 1, 6 / 12, 5 / 12]}>{children}</Col>
+)
+
+export const SecondColum = ({ fillWidth = false, children }) => (
```
jsms90: Typo SecondColum**n**

src/components/JoinUs/Learning.js
```diff
@@ -0,0 +1,62 @@
...
+  height: 100%;
+`
+
+const Learning = ({ data: { title, text, featuredInsights, list } }) => (
+  <Section greyBg>
+    <Row>
+      <FirstColumn>
+        <H2>{title}</H2>
+      </FirstColumn>
+      <SecondColum width={[1, 1, 1, 1, 6 / 12, 5 / 12]}>
```
jsms90: Based on [this line](https://github.com/yldio/yld.io/pull/80/files#diff-68a74b60a6cd09266fc4ff3341c7a014R19) in the definition of SecondColumn, should this be achieved by adding the `fullWidth` prop?

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,97 @@
...
+    return content
+  }
+
+  return <GreyBG static>{content}</GreyBG>
+}
+
+/**
+ * Elements
+ */
+export const Hr = styled.hr`
```
jsms90: We have an Hr component in the `Common` folder that you can use

src/components/JoinUs/Learning.js
```diff
@@ -0,0 +1,62 @@
...
+        <List list={list} />
+      </SecondColum>
+    </Row>
+    <Separator />
```
jsms90: I'm not sure that we need a new `Separator` component.

I think [the padding on Separator](https://github.com/yldio/yld.io/pull/80/files#diff-68a74b60a6cd09266fc4ff3341c7a014R90) belongs on the [`TitleAndList` component](https://trello.com/c/ycIBI03f/277-titleandlist-component) - i.e. on the `Row` here (line 27)

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,97 @@
...
+  margin: 18px 0;
+  opacity: 1;
+  background: ${props => props.theme.colors.grey};
+`
+
+export const LightH3 = styled(H3)`
```
jsms90: Do we need a new element? I think this is just `<H3 normal>`?

If we have a new colour variation, it might be best to add this as a prop to our existing `H3` component?

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,97 @@
...
+  font-weight: normal;
+  color: ${props => props.color || props.theme.colors.text};
+`
+
+export const ShiftedColumn = styled(Col)`
+  ${breakpoint('tablet')`
+    margin-left: 8.33333%;
+  `};
+`
+export const TitleAndBody = ({ title, body }) => (
```
jsms90: Worth moving to the `Common` folder, for use on other pages?

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,97 @@
...
+  <React.Fragment>
+    <Row>
+      <Col width={[1, 1, 1, 1, 6 / 12, 6 / 12, 5 / 12]}>
+        <LightH3>{title}</LightH3>
+      </Col>
+    </Row>
+    <Row>
+      <Col width={[1, 1, 1, 1, 6 / 12, 6 / 12, 5 / 12]}>
```
jsms90: Any need to repeat this twice & have two `Row`s? Could we not just put the H3 & Paragraph right next to each other?

src/components/JoinUs/Learning.js
```diff
@@ -0,0 +1,62 @@
...
+    <TitleAndBody
+      title="Some insights into personal growth at YLD"
+      body={text}
+    />
+    <CompensatedRow>
+      {featuredInsights.map((el, idx) => (
+        <CompensatedCol width={[1, 1, 1, 4 / 12]} key={idx}>
+          <InsightAnchor href={el.url}>
+            <Padding top={2} style={{ width: '100%' }}>
```
jsms90: Padding top = 36px (i.e. `3`) in Zeplin for desktop & large tablet:
image

src/components/JoinUs/Learning.js
```diff
@@ -0,0 +1,62 @@
...
+              <Margin bottom={1}>
```
jsms90: How come we have margin bottom on the image & padding top on the paragraph? Do we need this one? Maybe I'm missing something.

src/components/JoinUs/Learning.js
```diff
@@ -0,0 +1,62 @@
...
+                <Image image={el.image} />
+              </Margin>
+            </Padding>
+            <Padding top={1} bottom={1} style={{ flex: 1 }}>
```
jsms90: Space between image & link on Zeplin is `18px` for desktop & large tablet:
image

src/components/JoinUs/Learning.js
```diff
@@ -0,0 +1,62 @@
...
+              <Paragraph bold noMargin>
+                {el.image.title}
```
jsms90: Is it safe to assume that Greta will always ensure that the title of the image that she uploads starts with "YLD perspectives..."? i.e. that the title of the image that she uploads corresponds to the text we want to show on this page?

Perhaps we should use the "name" field in the content model for this:
image

I.e. name = "Tom Gallacher" insetad of "Featured insight 1"
image

We would then hard code "YLD perspectives: " into this line of the codebase?

src/components/JoinUs/ViewOpenPositions.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { Padding } from 'styled-components-spacing'
+import { Grid, Row, Col } from '../grid'
+import StyledLink from '../styledLink'
+import { H2 } from '../Typography'
+
+const ViewPositions = ({ text }) => (
+  <Grid>
+    <Padding
+      top={{ smallPhone: 4, smallTablet: 0 }}
```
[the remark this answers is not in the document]
antoniocapelo: It's a strange scenario - as there's no padding on the spec.

If we check the current distance from the text container to the top of the page we get around 132 (see `height` value on the next screenshot):

image


and if we compare it with the same distance on the small tablet zepplin file:

image


In the large tablet design, however, the distance increases:

image


So I'll keep set a `0.5` top padding for small tablet and set it to 4 on tablet, does that seem ok with you?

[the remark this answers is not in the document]
jsms90: I think we should just ask designers

src/components/JoinUs/Learning.js
antoniocapelo: you're right

src/components/JoinUs/ViewOpenPositions.js
```diff
@@ -0,0 +1,25 @@
...
+      bottom={{ smallPhone: 3.5, desktop: 5 }}
```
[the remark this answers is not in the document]
jsms90: Hmm, not really. What I was saying above was about spacing between the navbar & this component's title (i.e. top padding)

This comment is a separate and very minor tweak to the bottom padding. What you've got is right 3 out of 4 breakpoints (mobile, small tablet & desktop). But to make this conform to Zeplin, you're just missing `tablet: 5`.

Does that make sense?

src/components/JoinUs/Learning.js
antoniocapelo: I mean, not setting the prop

src/components/JoinUs/Learning.js
antoniocapelo: you're right, it's done now

src/components/JoinUs/Learning.js
antoniocapelo: In order to get those 18px distance between them (as you mentioned above), this is being done by setting a 6px margin bottom on the image and a 16px padding top on the paragraph. I guess this can be applied to different elements, but the end result's the same. Not sure what you mean

src/components/JoinUs/Learning.js
antoniocapelo: > `I.e. name = "Tom Gallacher" instead of "Featured insight 1"`

I was not sure how you normally deal with this situations - I used `Featured insight 1` because it seemed reasonable — wasn't so sure on naming the model itself with the title of the link since that could change in the future. 
From content editor perspective, I'd prefer to be able to change the title to whatever I want (if for some reason I want to remove the 'YLD Perspectives' part, I can. I also thought that it was pretty clear that the current title is in the format `YLD Perspectives: name` so I thought it made sense.

I can change it to have it hardcoded, don't have much strong feelings regarding this

src/components/JoinUs/Learning.js
antoniocapelo: From the ticket and designs, I don't feel like this needs to be on that component..what makes you think that?
(Also, if we decided that it does make part of the component, wouldn't this code be removed when that component is done?)

src/components/JoinUs/ViewOpenPositions.js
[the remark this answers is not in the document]
antoniocapelo: the `smallTablet: 0.5` was to add to the natural distance caused by the top menu height, in order to create the distance that we see on zepling — from the text container to the top of the window. Let's see what Antonas and Tom decide regarding this?

src/components/JoinUs/elements.js
antoniocapelo: faulty keyboard! :D 👍

src/components/JoinUs/elements.js
antoniocapelo: Maybe you're right

antoniocapelo: In terms of style, it looks like an `h5` - small title, I'll change to that

src/components/JoinUs/Work.js
```diff
@@ -0,0 +1,120 @@
...
+      {children}
+    </Padding>
+  </CaseStudyLink>
+)
+
+const CaseStudyWrapper = styled(Padding)`
+  display: flex;
+  width: 100%;
+`
+const CaseStudyTitle = styled(H3)`
```
jsms90: This should already be one of [our Typography components](https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5bf54a64af23577ecb48d2be)

...but to be fair, in terms of the codebase, I don't actually see an appropriate one :confused: 

image

src/components/JoinUs/Work.js
```diff
@@ -0,0 +1,120 @@
...
+    <Row>
+      <FirstColumn>
+        <H2>{title}</H2>
+      </FirstColumn>
+      <SecondColumn>
+        <List list={list} />
+      </SecondColumn>
+    </Row>
+    <Separator />
+    <TitleAndBody title="Some of our work" body={text} />
```
jsms90: Does Greta not want to be able to edit these titles in Contentful?

src/components/JoinUs/Work.js
```diff
@@ -0,0 +1,120 @@
...
+  flex: 1;
+`
+
+const CaseStudy = ({ children, ...props }) => (
+  <CaseStudyLink {...props}>
+    <Padding
+      vertical={{ smallPhone: 2, smallTablet: 1.5, tablet: 3 }}
+      horizontal={{ smallPhone: 2, smallTablet: 1.5, tablet: 2 }}
+    >
+      {children}
```
jsms90: Since CaseStudy isn't being re-used anywhere, perhaps it's even clearer and more readable to pop the children (i.e. title and paragraph) directly in here?

src/components/JoinUs/Work.js
```diff
@@ -0,0 +1,120 @@
...
+      </FirstColumn>
+      <SecondColumn>
+        <List list={list} />
+      </SecondColumn>
+    </Row>
+    <Separator />
+    <TitleAndBody title="Some of our work" body={text} />
+    <CompensatedRow>
+      {someWork.slice(0, MAX_CASE_STUDIES).map((cs, idx) => {
+        const lightText = getColorLuminance(cs.posterColor) < 127.5
```
jsms90: Cool :+1: :smile:

src/components/JoinUs/elements.js
antoniocapelo: I think that initially they had different widths for different breakpoints, then the design must have changed. Now it doesn't make any sense, we can have them in the same col

src/components/JoinUs/elements.js
antoniocapelo: Yeah probably. Would a `dark` prop do?

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,89 @@
...
+      source={list}
+    />
+  </Padding>
+)
+
+export const Section = ({ children, greyBg, ...props }) => {
+  const content = (
+    <Grid {...props}>
+      <Padding
+        top={{ smallPhone: 3, tablet: 4 }}
```
jsms90: Asked Tom in person, what this rule is supposed to be. I am told:
- desktop & large tablet - top: 24, bottom: 36, left: 24, right: 24
- small tablet: top: 18, bottom: 24, left: 18, right: 18
- mobile: top: 24, bottom: 24, left: 24, right: 24

Note: we're only gonna be able to change this once [the typography components get sorted out](https://trello.com/c/PRcbDKOg/247-please-align-typographic-padding-values-with-the-design-documentation-including-top-and-bottom-padding)

src/components/JoinUs/Work.js
antoniocapelo: Yup, not sure what to do here. The specs for the typography vs what we have are hard to match, at least for me..

src/components/JoinUs/Work.js
```diff
@@ -0,0 +1,120 @@
...
+
+  ${breakpoint('tablet')`
+    font-size: ${remcalc(26)};
+    line-height: ${remcalc(30)};
+  `};
+`
+
+const Work = ({ data: { title, list, text, someWork } }) => (
+  <Section>
+    <Row>
```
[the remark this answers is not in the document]
antoniocapelo: This is one of the cases that we already discussed — a perfect candidate to extract. Wasn't sure if this was the best time to do it, but I can do it if needed

src/components/JoinUs/Learning.js
jsms90: Yeah that's fine about the hard coding. You're totally right about this:
> I'd prefer to be able to change the title to whatever I want (if for some reason I want to remove the 'YLD Perspectives' part, it's possible without a release).

I guess my main point was that it might be better to use a field in Contentful, that Greta explicitly has to type in, rather than grabbing the image's title?

src/components/JoinUs/Work.js
antoniocapelo: I really don't know. 
When talking with Sara, I understood that these main titles — if provided and approved by Fin/Greta — wouldn't change. 
Does every text field need to be customizable?

src/components/JoinUs/Work.js
```diff
@@ -0,0 +1,120 @@
...
+      <FirstColumn>
+        <H2>{title}</H2>
+      </FirstColumn>
+      <SecondColum width={[1, 1, 1, 1, 6 / 12, 5 / 12]}>
+        <List list={list} />
+      </SecondColum>
+    </Row>
+    <Separator />
+    <TitleAndBody title="Some of our work" body={text} />
+    <CompensatedRow>
```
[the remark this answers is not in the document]
antoniocapelo: You're suggesting creating a component for a 3 column layout? (not sure if I understood)

src/components/JoinUs/Work.js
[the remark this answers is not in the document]
jsms90: Up to you. We can push the page in and refactor afterwards if you prefer.

It's usually just harder to convince product that refactoring is necessary (after the feature is already live), than it is to bake componentisation into the making of a page :woman_shrugging:

src/components/JoinUs/elements.js
antoniocapelo: I think it's best to analyse the Trello ticket and check if the name makes sense, if there are variations and then extract. do you agree?

src/components/JoinUs/elements.js
antoniocapelo: So, is there any action to be taken regarding this file, for the time being?

src/components/grid.js
```diff
@@ -163,3 +163,9 @@ export const Row = props => (
 export const Col = props => (
   <Flex block px={[1, 3, 60, 42, 48, 48]} {...props} />
 )
+
+export const CompensatedRow = props => <Flex mx={[-1, -1.5, -2]} {...props} />
+
+export const CompensatedCol = props => (
+  <Flex block px={[1, 1.5, 2]} {...props} />
```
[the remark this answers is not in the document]
antoniocapelo: I don't think is a problem created with the newly added code. I've mentioned this 'double-gutter' problem with the current `Col` and `Row` implementation.

The case you're referring, the `TitleAndBody` uses regular `Col` component which indeed has that problem. This `Compensated` components were created exactly to address this issue..
When openning this MR I wrote a couple of comments, and one was:

> Created compensated cols to better adapt to current design specs (gutter width)
> 
> Our design specs show our gutter width at around 48px but our current Col components use a 48px each, which causes our grid is visually the double of what it needs (48+48).
> I don't know if this is a known issue or not so until we can establish a convention, I created the CompensatedCol and CompensatedRow (probably not the best names, but I'm hoping they're temporary) that allow for the single gutter width."

src/components/JoinUs/ViewOpenPositions.js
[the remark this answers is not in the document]
jsms90: I don't understand. `smallTablet: 0.5`? Where's that?

Maybe we should just jump on a call tomorrow?

src/components/JoinUs/elements.js
jsms90: You're totally right, we shouldn't skip from h2 to h5. It should be `as="hX"` where X is the next number down in the semantic order (for accessibility reasons)

I think it would be helpful if we called our typographic components after the designers' names, rather than being tied to the semantic names.

src/components/JoinUs/elements.js
jsms90: :laughing:

src/components/JoinUs/elements.js
antoniocapelo: ok!

antoniocapelo: done

src/components/JoinUs/Work.js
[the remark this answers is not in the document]
jsms90: Maybe we can revisit this idea as part of the post-merge componentisation. Might not even work. I haven't thought it through. Just an idea

src/components/JoinUs/ViewOpenPositions.js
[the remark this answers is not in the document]
antoniocapelo: probably! ping me when your available

src/components/grid.js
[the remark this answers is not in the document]
jsms90: Yeah you're right. The existing `Col` creates this problem too. But I don't think the new components fix the issue.

It still uses `px` - i.e. padding on both sides. So whenever you have columns next to each other, there will still be padding on the right of one & padding on the left of the one next to it...

Am I missing something?

src/components/JoinUs/Work.js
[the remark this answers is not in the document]
antoniocapelo: Ok, I just didn't went down that road because it's a fairly simple scenario (one row, 3 cols), not sure if a component makes sense, but yeah we can discuss it afterwards

src/components/JoinUs/Work.js
jsms90: Hmm, yeah ok. Maybe worth a review tomorrow. But I reckon the right step would be for us to adjust the existing typography components to match designs, rather than make an exception for this case.

What do you think?

src/components/JoinUs/Learning.js
antoniocapelo: Makes sense, probably using a dedicated field is a best bet

src/components/JoinUs/Work.js
jsms90: Yeah I get where you're coming from. This does seem to be the existing pattern across a lot of our current pages. But the first time Sara & Greta were going through how to use Contentful, I remember her wanting to change the title for the Events section on the homepage & couldn't because it was hard coded.

Just went over to ask her in person. She said it would ideally be editable.

In an ideal world, I reckon these steps would be done before a ticket was considered 'ready' for engineering:
image

whereby engineering & product would liase closely, when creating the content model for new pages.

But I doubt that was possible in the past, while the focus was on shipping as fast as possible & the team had really tight deadlines, to get things done before [DSL](https://www.designsystemslondon.com/)

src/components/JoinUs/elements.js
jsms90: Depends whether designers are happy pushing this page live before the typography components are fixed

src/components/grid.js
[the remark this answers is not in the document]
antoniocapelo: Yes, there's padding on the left and right of two adjacent cols, but that padding is half the gutter, so when summed up, the gutter should stay visually correct.
Does that makes sense?

src/components/JoinUs/Work.js
antoniocapelo: I think so, yes. Let's have a quick chat before I start implementing to make sure I'll go in the right direction. (or maybe someone else wants to pick it up)

src/components/JoinUs/Work.js
antoniocapelo: 😄

src/components/JoinUs/elements.js
antoniocapelo: I'm guessing that this `Section` component does not depend on the typography, right? It's just an aux component to replicate the paddings set on the designs? Or am I looking at the wrong thing?

src/components/JoinUs/Learning.js
antoniocapelo: That call really help understand this (I thought you were saying that the Hr belong on the `TitleAndList` comp 😃 
I've just created a `TitleAndList` component that can be extracted in the future, that applies that correct padding. I'm not extracting it already since it depends on those `Compensated` grid elements - but at least I think part of the job of creating that compo is done :)

src/components/JoinUs/Learning.js
antoniocapelo: You're right, we have a measure of 18 px on our theme, it's much quicker to use that as a padding on the `p`. It's done now :)

src/components/JoinUs/OpenSource.js
```diff
@@ -0,0 +1,99 @@
...
+`
+
+const Talks = ({ data }) => (
+  <CompensatedRow>
+    {data.slice(0, MAX_VIDEOS).map(({ title, link }, idx) => (
+      <TalkColumn width={[1, 1, 1, 6 / 12, 6 / 12, 4 / 12]} key={idx}>
+        <Padding top={2}>
+          <TalkLink px={2} py={2} href={link} target="_blank">
+            <Padding horizontal={2} vertical={2}>
+              <Margin bottom={1}>
```
jsms90: What aspect of the designs does this achieve?

src/components/JoinUs/OpenSource.js
```diff
@@ -0,0 +1,99 @@
...
+  `}
+`
+
+const Talks = ({ data }) => (
+  <CompensatedRow>
+    {data.slice(0, MAX_VIDEOS).map(({ title, link }, idx) => (
+      <TalkColumn width={[1, 1, 1, 6 / 12, 6 / 12, 4 / 12]} key={idx}>
+        <Padding top={2}>
+          <TalkLink px={2} py={2} href={link} target="_blank">
+            <Padding horizontal={2} vertical={2}>
```
jsms90: According to Zeplin, the CompactVideoLink component should have this padding around it:
image

Where does `24px` (i.e. `horizontal={2} vertical={2}`) come from?

jsms90: Hmm.... I was looking at `CompactVideoLink` because that's the trello ticket you picked up. But, looking at the page-level designs in Zeplin, it looks like we're actually using `StandAloneVideoLink` on this page?

In which case, these are the appropriate paddings:
image

So we would have something like:
```
const StandAloneVideoLink = props => (
  <Wrapper> // i.e. layout with compensated row, columns, etc.
    <Padding
      vertical={1.5} // i.e. 18px
      left={14} // since 14px isn't a value in our spacing sizes
      right={2} // back to spacing values
    >
      <Padding vertical={57}>
        <CompactVideoLink />
      </Padding>
    </Padding>
  </Wrapper>
)
```

Or you could amalgamate the 2 `Padding` components that are sitting next to each other?

src/components/JoinUs/ViewOpenPositions.js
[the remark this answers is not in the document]
antoniocapelo: Was just checking, the padding on small tablet is 54px === 3.5, right?

image

src/components/JoinUs/OpenSource.js
```diff
@@ -0,0 +1,99 @@
...
+    &:last-child {
+      display: flex;
+    }
+  `}
+`
+
+const Talks = ({ data }) => (
+  <CompensatedRow>
+    {data.slice(0, MAX_VIDEOS).map(({ title, link }, idx) => (
+      <TalkColumn width={[1, 1, 1, 6 / 12, 6 / 12, 4 / 12]} key={idx}>
```
jsms90: Would be good to extract everything from this point downwards into a common component called StandAloneVideoLink, so that it can be used in other pages

src/components/JoinUs/Work.js
antoniocapelo: Just mentioned this on [this trello card](https://trello.com/c/PRcbDKOg) so it's not forgotten

src/components/JoinUs/ViewOpenPositions.js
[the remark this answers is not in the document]
jsms90: Sorry! Screenshot above is from large tablet, not small tablet.

My bad :sweat:

src/components/JoinUs/elements.js
antoniocapelo: I'll mention it on the trello ticket (https://trello.com/c/wWqav3AO/295-displaybodybody-component)

jsms90: I think it's safe to move this component into the Common folder within this PR, rather than leaving it in `JoinUs/elements.js`?

But, yeah, I agree that it's in the scope of [this ticket](https://trello.com/c/ycIBI03f/277-titleandlist-component), to ensure consistency by using what you have created everywhere else on the website.

src/components/JoinUs/Work.js
```diff
@@ -0,0 +1,106 @@
...
+  return colorLuminance(rgbValue.r, rgbValue.g, rgbValue.b)
+}
+
+const CaseStudyLink = styled(Link)`
+  background: ${props => props.bg};
+  color: ${props => props.color || props.theme.colors.white};
+  flex: 1;
+`
+
+const CaseStudy = ({ title, lightText, ...props }) => (
```
jsms90: Asked Tom & we've agreed that this component needs its own Zeplin documentation. Especially because Beatriz is looking to use this on another page. So it might be useful to move this out into the Common folder now

src/components/JoinUs/Work.js
```diff
@@ -0,0 +1,106 @@
...
+  <CaseStudyLink {...props}>
+    <Padding
+      vertical={{ smallPhone: 2, smallTablet: 1.5, tablet: 3 }}
+      horizontal={{ smallPhone: 2, smallTablet: 1.5, tablet: 2 }}
```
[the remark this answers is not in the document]
jsms90: Mentioned in [this new trello ticket](https://trello.com/c/EOTTpbx6/323-another-casestudy-component-needs-its-own-distinctive-name)

src/components/JoinUs/Work.js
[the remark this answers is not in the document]
jsms90: From our call this morning, I think you decided to do this work within this PR rather than us raising a new trello ticket for this component. Right?

src/components/grid.js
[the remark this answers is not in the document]
antoniocapelo: I'm closing this as per we discussed on our slack call. This may turn into the 'default' col solution, let's just get a GO from the design team for fixing the rest of the 'double-gutter' situations

src/components/JoinUs/OpenSource.js
```diff
@@ -0,0 +1,111 @@
...
+  List,
+  TitleAndBody
+} from './elements'
+
+const MAX_VIDEOS = 3
+
+const getThumbnail = videoId =>
+  `https://img.youtube.com/vi/${videoId}/maxresdefault.jpg`
+
+const TalkLink = styled.a`
```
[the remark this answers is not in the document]
antoniocapelo: This whole `anchor` will be replaced with the  'soon to be created' StandaloneVideoLink... So I don't really know if it's worth it..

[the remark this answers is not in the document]
antoniocapelo: this a `display: block` item, so I don't see how they'd fit inside a paragraph. This is basically the anchor with a border, used on each talk link. But this will changed and will be part of the `StandaloneVideoLink`, that's yet to be developed.

Don't think there's much to do here, then, what do you think?

src/components/JoinUs/OpenSource.js
```diff
@@ -0,0 +1,99 @@
...
+  border: 1px solid #eeefee;
+  display: block;
+  flex: 1;
+`
+const TalkTitle = styled.div`
```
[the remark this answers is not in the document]
antoniocapelo: agree, changed to `FlexContainer`

src/components/JoinUs/OpenSource.js
antoniocapelo: the `StandAloneVideoLink` is different — the design for these links have changed — and from what I understand, it has to be built on top of the `CompactVideoLink`. Currently, this link does not match with the design

src/components/JoinUs/Work.js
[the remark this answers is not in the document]
antoniocapelo: Yes!

src/components/JoinUs/ViewOpenPositions.js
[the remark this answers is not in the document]
antoniocapelo: @jsms90 I've talked with antonas and he told me to follow the padding rules for the 'About Us' page (https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5c112c2ac5dd4c508a0eaadc). And based on that, the paddings for this components are correct. The visual discrepancy we see on the top spacing, when comparing with the designs, it's because our header has wrong paddings coded. I've created a ticket addressing that and have put it on our 'Ready for Engineering' list - https://trello.com/c/FoKYTdV9

src/components/JoinUs/OpenSource.js
```diff
@@ -0,0 +1,99 @@
...
+  display: flex;
+`
+const PlayIcon = styled.img`
+  min-height: ${remcalc(36)};
+  max-width: ${remcalc(36)};
+`
+const TruncatedParagraph = styled(Paragraph)`
+  max-height: 3rem;
```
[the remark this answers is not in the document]
jsms90: > remcalc outputs rem units, so I don't understand why should I do it?

Purely for consistency. It's true that they both become rem in the browser. But in our codebase, we're now mixing methods.

> it's calculated based on our current font-size for this element

So is this 2 * line height? What's the restriction rule supposed to be, according to designs? Looking at Zeplin, I'm not even sure

src/components/JoinUs/OpenSource.js
jsms90: I don't really get it :disappointed: 

This is from the Join Us page on Zeplin:
image

which is definitely making use of the StandAloneVideoLink component:
image

Basing the StandAloneVideoLink on the CompactVideoLink makes sense :+1:

But either way, the component being used on this page is definitely the StandAloneVideoLink component, no? :confused:

src/components/JoinUs/OpenSource.js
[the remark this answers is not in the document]
antoniocapelo: closing this, hopefully clarified on the `StandaloneVideoLink`

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
+import ReactMarkdown from 'react-markdown'
+import styled from 'styled-components'
+import { Padding } from 'styled-components-spacing'
+import breakpoint from 'styled-components-breakpoint'
+import { Grid, Row, Col } from '../grid'
+import { H2, H3, H5, Paragraph } from '../Typography'
+import GreyBG from '../GreyBG'
```
jsms90: Can I suggest that you use `GreyBackgroundWithoutOffset` for now, until we pick up [this ticket](https://trello.com/c/N7vocsNh/268-combine-greybg-greybackgroundwithoutoffset).

I think that using this component is what is causing some incorrect spacings between white and grey sections.

This isn't suuuuper easy to see in a screenshot but have a look [on the preview](https://deploy-preview-80--yldio.netlify.com/join-us/):
- between OpenSource & Community sections
image
- space between title & top of grey background on OpenPositions

antoniocapelo: Check again please

src/components/JoinUs/Perks.js
```diff
@@ -0,0 +1,79 @@
...
+import { Paragraph, H2 } from '../Typography'
+import Image from '../Common/Image'
+
+import { Section, ShiftedColumn } from './elements'
+
+const PerksListContainer = styled(ShiftedColumn)`
+  display: grid;
+  row-gap: ${remcalc(24)};
+  grid-template-columns: 1fr 1fr;
+  column-gap: 42px;
```
jsms90: If you compare the Zeplin version of the Join Us page:
image

to coded:
image

you can see that this spacing is not quite as designed

### See Grid documentation in Zeplin
image

image

It's the gutters that are hard-coded to a specific value & then the column widths are dependent on what is left over.

etc. (More breakpoints visible in Zeplin)

---

As much as I love the idea of using CSS Grid!! :heart_eyes:, I think it's worth us deciding on common conventions to use throughout the codebase. Since we don't use CSS Grid anywhere else to achieve our columns & rows, I'm not sure that it's a great idea to be mixing solutions?

To be honest, I think it would make our lives muuuuch easier if we moved to CSS Grid everywhere, but that's a big refactor job & I think consistency is more important than _which_ solution you go with.

But I'm totally open to being wrong on this.

src/components/JoinUs/OpenSource.js
antoniocapelo: It is, but at the time of this discussion, the `StandAloneVideoLink` wasn't yet done. Now it's done and used appropriately

src/components/JoinUs/Perks.js
```diff
@@ -0,0 +1,79 @@
...
+  padding: ${remcalc(12)} 0;
+`
+const Perk = ({ icon, description }) => (
+  <PerkContainer>
+    <Icon image={icon} />
+    <PerkDescription noMargin>{description}</PerkDescription>
+  </PerkContainer>
+)
+
+const PerksList = ({ perks }) => (
```
jsms90: I think we should avoid naming things `List` unless they are semantically lists under the hood?

Arguably, perhaps these could/should be a list?

src/components/JoinUs/Perks.js
```diff
@@ -0,0 +1,79 @@
...
+    grid-template-columns: 1fr 1fr;
+    row-gap: ${remcalc(36)};
+    column-gap: 42px;
+    padding-top: ${remcalc(12)};
+  `};
+`
+
+const Icon = styled(Image)`
+  width: 60px;
+  height: 60px;
```
jsms90: I think I'm right in saying that we currently have multiple ways to achieve this restricted size across our codebase right now.

Perhaps it's worth us add pattern, of specifying/restricting the width & height, via a prop on Image. What do you think?

Could also be the remit of a separate ticket. Just throwing it out there.

src/components/JoinUs/Perks.js
```diff
@@ -0,0 +1,79 @@
...
+  `};
+
+  ${breakpoint('smallTablet')`
+    grid-template-columns: 1fr 1fr 1fr;
+  `};
+
+  ${breakpoint('tablet')`
+    grid-template-columns: 1fr 1fr;
+    row-gap: ${remcalc(36)};
+    column-gap: 42px;
```
jsms90: If you compare the Zeplin version of the Join Us page:
### Zeplin
image

### Coded
image

you can see that this spacing is not quite as designed (48px on desktop, not 42px)

### See Grid documentation in Zeplin
image

image

It's the gutters that are hard-coded to a specific value & then the column widths are dependent on what is left over.

etc. (More breakpoints visible in Zeplin)

src/components/JoinUs/Perks.js
```diff
@@ -0,0 +1,79 @@
...
+import breakpoint from 'styled-components-breakpoint'
+import { Row, Col } from '../grid'
+import { Padding } from 'styled-components-spacing'
+import { Paragraph, H2 } from '../Typography'
+import Image from '../Common/Image'
+
+import { Section, ShiftedColumn } from './elements'
+
+const PerksListContainer = styled(ShiftedColumn)`
+  display: grid;
```
jsms90: As much as I love the idea of using CSS Grid!! :heart_eyes:, I think it's worth us deciding on common conventions to use throughout the codebase. Since we don't use CSS Grid anywhere else to achieve our columns & rows, I'm not sure that it's a great idea to be mixing solutions?

To be honest, I think it would make our lives muuuuch easier if we moved to CSS Grid everywhere, but that's a big refactor job & I think consistency is more important than _which_ solution you go with.

But I'm totally open to ideas, if you think it's worth deviating from our conventions for this scenario alone?

src/components/JoinUs/Perks.js
```diff
@@ -0,0 +1,79 @@
...
+    ))}
+  </PerksListContainer>
+)
+
+const Perks = ({ data: { title, text, perks } }) => (
+  <Section>
+    <Row>
+      <Col width={[1, 1, 1, 1, 1, 5 / 12]}>
+        <H2>{title}</H2>
+        <Paragraph muted>{text}</Paragraph>
```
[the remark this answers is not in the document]
antoniocapelo: image

It varies, sometimes it's 23 other 20, probably 24px is a better value since it complies with our spacing system.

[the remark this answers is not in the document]
antoniocapelo: then again, 8px is the default for the h2, I'll add the 12px on the body text

src/components/JoinUs/Perks.js
antoniocapelo: imo CSS grid was the adequate solution for this, regarding the control of the spacings and horizontal/vertical alignment. I don't think it should be used in replacement of the current row/column solution, but as a complement.

Besides Grid having a great browser coverage, before using it, I searched for usages of `grid` in the project and found at list to, so I considered it as a valid solution for the current conventions.

src/components/JoinUs/OpenPositions.js
```diff
@@ -0,0 +1,96 @@
...
+            {commitment}
+          </Paragraph>
+        </Padding>
+      </JobLink>
+      <Line />
+    </JobLinkWrapper>
+  </CompensatedCol>
+)
+
+const DirectApplication = ({ title, text }) => (
```
jsms90: Can we use the existing `TalkToUs` component that's in Common?

src/components/JoinUs/OpenPositions.js
```diff
@@ -0,0 +1,96 @@
...
+    </Row>
+  </Padding>
+)
+
+class OpenPositions extends React.PureComponent {
+  render() {
+    const { title, directApplicationTitle, text } = this.props.data
+
+    return (
+      <Section greyBg id="open-positions">
```
jsms90: Is this id being used for something?

src/components/JoinUs/Perks.js
```diff
@@ -0,0 +1,79 @@
...
+import remcalc from 'remcalc'
+import breakpoint from 'styled-components-breakpoint'
+import { Row, Col } from '../grid'
+import { Padding } from 'styled-components-spacing'
+import { Paragraph, H2 } from '../Typography'
+import Image from '../Common/Image'
+
+import { Section, ShiftedColumn } from './elements'
+
+const PerksListContainer = styled(ShiftedColumn)`
```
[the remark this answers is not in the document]
antoniocapelo: That's a good approach, I've updated the code

src/components/JoinUs/elements.js
```diff
@@ -0,0 +1,85 @@
...
+      renderers={{
+        heading: props => <H5 as="h3" noMargin bold {...props} />, // eslint-disable-line react/display-name
+        paragraph: props => <Paragraph muted {...props} /> // eslint-disable-line react/display-name
+      }}
+      source={list}
+    />
+  </Padding>
+)
+
+export const Section = ({ children, greyBg, ...props }) => {
```
jsms90: This is nice :+1: I reckon we should find a consistent way to do this across the codebase?

If we're going to call it `Section`, that's a pretty good indication that it should be an HTML5 `section` under the hood. I reckon this would actually be a good idea across the whole website, by [changing `Grid` to always be section, instead of a div](https://github.com/yldio/yld.io/blob/master/src/components/grid.js#L110)?

Similarly, I would probably also pop a `backgroundColor` prop onto the Grid component (as opposed to just `greyBg` here). Maybe we could then create a utility function that applies the appropriate colour from the theme, as the background colour, & use this pattern in our `Header` (aka NavBar) as well?

src/components/JoinUs/OpenPositions.js
```diff
@@ -0,0 +1,96 @@
...
+import { FirstColumn, SecondColumn, Section } from './elements'
+import ExternalAnchor from '../Common/ExternalAnchor'
+import Hr from '../Common/Hr'
+import StyledLink from '../styledLink'
+
+const Line = styled(Hr)`
+  width: ${remcalc(60)};
+`
+
+const JobLinkWrapper = styled.div`
```
jsms90: We have `styled-flex-component` that we can use instead of a `div` + `display: flex`.

Again, not because either is necessarily better or worse. Just for consistency across our codebase.

src/components/JoinUs/Perks.js
```diff
@@ -0,0 +1,79 @@
...
+  <PerkContainer>
+    <Icon image={icon} />
+    <PerkDescription noMargin>{description}</PerkDescription>
+  </PerkContainer>
+)
+
+const PerksList = ({ perks }) => (
+  <PerksListContainer width={[1, 1, 1, 1, 1, 5 / 12]}>
+    {perks.map((perk, idx) => (
+      <Perk
```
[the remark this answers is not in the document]
antoniocapelo: Yes, antonas and tom showed that they have a common component prepared (not sure if fully documented) for this item

src/components/JoinUs/Perks.js
antoniocapelo: not sure what you meant in this comment - seems like a mixture of a new and an old comment? 

Regarding the gutters/column spacing - this is the bug that we've been talking: our whole grid/column system is not 100% accurate to the designs. I don' think it should be fixed in this MR. (If that's not what you were referring, sorry :) )

src/components/JoinUs/Perks.js
jsms90: Fair enough.

You're probably right. Given that this is a component which requires spacing along both axes, CSS Grid is the right solution.

What I'm commenting on is probably more that [we need documentation](https://trello.com/c/DrXHIMWe/207-document-how-our-grid-works-should-be-used-in-combination-with-the-other-style-libraries-we-use) actually

src/components/JoinUs/Perks.js
antoniocapelo: You're right, they should be a list

antoniocapelo: done! 👍

src/components/JoinUs/Perks.js
antoniocapelo: @jsms90 sure I think it makes sense, should cover most of the solutions.

src/components/JoinUs/OpenPositions.js
```diff
@@ -0,0 +1,96 @@
...
+        </Padding>
+      </Section>
+    )
+  }
+
+  renderJobsForlocation = (jobs, location, key) => {
+    return (
+      <Padding top={3} key={key}>
+        <H5 as="h3">{location}</H5>
+        <Padding top={1}>
```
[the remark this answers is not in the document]
antoniocapelo: indeed. done!

src/components/JoinUs/OpenPositions.js
antoniocapelo: yes, for opening the page directly in this section (if we open /join-us#open-positions) and for directing the user here if the `View Open Positions` link in the top section is clicked

src/components/JoinUs/OpenPositions.js
antoniocapelo: where's that component?

src/components/JoinUs/elements.js
antoniocapelo: I called it section because it's inside the 'scope' of the page, so it's a section of the page. If this was to be moved for a `common` folder, then yes it should reflect the tag element name (or letting it pass an `as` prop). 

Regarding changing grid to always be a section, I'm not really sure if it will always make sense semantically (if for some reason we need inner grids - don't know if this will ever happen though). Our `background` components already render a `section` btw.

If this was/will be a common component, then yes a backgroundColor might make sense — for the purpose of this page, a simple bool was enough

src/components/JoinUs/OpenPositions.js
antoniocapelo: 👍

src/components/JoinUs/OpenPositions.js
jsms90: ah gotcha!

src/components/JoinUs/OpenPositions.js
```diff
@@ -0,0 +1,96 @@
...
+import JobsByLocation from '../jobsByLocation'
+import { Padding } from 'styled-components-spacing'
+import { Paragraph, H2, H5 } from '../Typography'
+import { FirstColumn, SecondColumn, Section } from './elements'
+import ExternalAnchor from '../Common/ExternalAnchor'
+import Hr from '../Common/Hr'
+import StyledLink from '../styledLink'
+
+const Line = styled(Hr)`
+  width: ${remcalc(60)};
```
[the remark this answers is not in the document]
antoniocapelo: Now I can, since it's been changed later on :D

src/components/JoinUs/Perks.js
antoniocapelo: the whole grid should respect that spacing. I didn't do since I though the grid/gutter was to be reviewed. But I now changed this section to follow the specs

jsms90:
Just a reminder to mention this to Greta before we go live - the way jobs are referred to in Lever make this component look a bit weird (the repetition of Contract):

image

PR #81, opened 2019-01-14
fix: fix unnecessary client check on header

antoniocapelo:

Currently we were checking if we were on the client side because we were using the `window.location.pathname` to do some logic. However, Gatsby injects the location as a prop on every page component. This way, we can pass our location to the `Layout` and from then to the `Header` component and use it safely with less checks.

I wish there was a way to also inject the location into the header directly but it doesn't seem like it.

Note: I ran `gatsby build` and served the build — everything looked ok.

Thoughts are welcome :)

## Review template

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -11,8 +11,6 @@ import menu from '../../images/menu.svg'
 import close from '../../images/menu_close.svg'
 import { MobileMenu, HomeLink, Close, DesktopMenu } from './elements.js'
 
-const isClient = () => typeof window !== 'undefined'
-
 const HiddenText = styled.h1`
   position: absolute;
   top: -9999px;
@@ -25,6 +23,8 @@ class Header extends Component {
   toggleMenu = () => this.setState(state => ({ menuOpen: !state.menuOpen }))
 
   render() {
+    const { location } = this.props
+
     return (
       <Grid>
         <Row>
@@ -33,10 +33,9 @@ class Header extends Component {
               <Padding top={2} bottom={3}>
                 <Flex alignCenter wrap justifyBetween>
                   <Link to="/">
-                    {isClient() &&
-                    !window.location.pathname.includes('engineering') &&
+                    {!location.pathname.includes('engineering') &&
                     !this.props.blue &&
-                    !window.location.pathname.includes('design') ? (
+                    !location.pathname.includes('design') ? (
                       <img
                         role="link"
                         tab-index="0"
@@ -46,8 +45,7 @@ class Header extends Component {
                       />
                     ) : null}
 
-                    {(isClient() &&
-                      window.location.pathname.includes('engineering')) ||
+                    {location.pathname.includes('engineering') ||
                     this.props.blue ? (
                       <Fragment>
                         <HiddenText>engineering</HiddenText>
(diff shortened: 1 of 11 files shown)


antoniocapelo:
If we were already using `PropTypes`, we could make the location prop as `required` for the `Header`. This would  prevent possible errors — if someone forgets to pass the location into the layout, from a new page. 

When running the build, however, gatsby will warn us about it, so I think we can live with it

src/components/Header/index.js
```diff
@@ -112,4 +110,10 @@ class Header extends Component {
   }
 }
 
+Header.defaultProps = {
```
jsms90: I believe `.defaultProps` comes directly from [PropTypes](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values). So unless we're importing the library, I reckon this is basically the same as comments, no?

src/components/Header/index.js
```diff
@@ -25,6 +23,8 @@ class Header extends Component {
   toggleMenu = () => this.setState(state => ({ menuOpen: !state.menuOpen }))
 
   render() {
+    const { location } = this.props
```
jsms90: If we want to force this value to default to an empty string, we could put this here?
```
const { location = '' } = this.props
```

src/pages/contact.js
```diff
@@ -82,7 +82,7 @@ class ContactUs extends Component {
     const site = this.props.data.site
     const page = this.props.data.allContentfulPage.edges[0].node
     return (
-      <Layout>
+      <Layout location={this.props.location}>
```
jsms90: Maybe destructure this from props in the lines above, before passing it down?

src/pages/index.js
```diff
@@ -15,60 +15,66 @@ import Services from '../components/Homepage/services'
...
+        <html lang="en" />
+      </Helmet>
       <Grid>
-        <Padding top={{ smallPhone: 2 }} />
-        <Padding top={{ smallPhone: 4, smallTablet: 5, desktop: 6 }} bottom={3}>
-          <SEOText text={content.seoText.content[0].content} />
-          <Padding bottom={{ smallPhone: 2, smallTablet: 4, desktop: 4 }} />
-          <Companies companies={content.companies} />
+        <Padding bottom={{ smallPhone: 0, smallTablet: 2, desktop: 2 }}>
+          <CaseStudy
```
jsms90: A minor thing, but it might be good to mention changes like this in the PR description? Just so a reviewer knows to check whether anything has changed unintentionally

src/components/Header/index.js
antoniocapelo: Not exactly, defaultProps is built in React components, which are props used as fallback when the parent-provided props are undefined, see https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values. 

The Proptypes lib allows for typechecking, which would be useful for warning the dev if a required value is missing, amongst other things. 

Does this make sense to you?

src/components/Header/index.js
antoniocapelo: That is what the defaultProps does 🙂

src/components/Header/index.js
jsms90: Ah yeah. That's the same file that I linked to. But I saw this and misunderstood:
> React.PropTypes has moved into a different package since React v15.5. Please use the prop-types library instead.

`.defaultProps` is still built into React? :+1:

src/components/Header/index.js
jsms90: Yeah. Only mentioned cos of my misunderstanding above :sweat_smile:

src/pages/contact.js
jsms90: Do you want me to push up this little thing?

Then I reckon merge?

antoniocapelo: Sure, I’m not in computer atm, if you can do it, go for it :)

src/components/Header/index.js
antoniocapelo: Yes ;)

PR #82, opened 2019-01-15
remove stackable case study from codebase

jsms90:
## Bug: master branch deployment failing because Stackable case study isn't being kept up to date
image

My bad for not keeping it up to date. But at standup we discussed how soon (if ever) we're likely to get permission to push this live.

Looks like not. Agreed to remooooove.


SaraVieira:
The canon one is also failing now :(

jsms90:
Ok. Found the problem - it's our content model

### Generic Case Study (aka 'design case study', or more accurately 'non-templated case study')
Has 'related case study' field:
image

Validation - assumed to be of type 'case study' only (i.e. design case studies can only link to other design case studies):
image

### Case Study (aka 'engineering case study', or more accurately 'templated case study')
Has 'related case study' field:
image

Validation - allowed to be of either type:
image

Based on latest product requirements, I'm going to change the content model so that we allow design case studies to link to either design or engineering case studies.

Note: this change needs to be reverted if/when product requirements change back. [See trello ticket](https://trello.com/c/pYQhjXba/226-define-related-case-study-for-current-case-study-pages)

SaraVieira:
I am super okay with this change

It wasn't like that because there weren't a lot of case studies

jsms90:
I believe that changing the content model + [this `gatsby-source-contentful` issue](https://github.com/gatsbyjs/gatsby/issues/3535#issuecomment-390048973) is responsible for the failing build & deployment.

For now, I have changed the content model back to what it was - i.e. design case studies can only be linked to engineering case studies
image

which also meant changing Greta's alteration to the content, so that Joyent's related case study is Trainline again (rather than Canon).

I suggest that we [spend a bit of time reviewing the content model](https://trello.com/c/d5MoLWHS/257-review-of-content-model), because this isn't the only situation where we're encountering discrepancies.

PR #84, opened 2019-01-18
remove unused code

SaraVieira:
Closes

https://trello.com/c/oWYKiekC/284-fix-issues-that-greta-was-having-with-entering-info-from-seo-copy-documents-into-contentful

and 

https://trello.com/c/kFZ3rLFQ/285-re-order-the-case-study-cards-in-the-seo-page-so-the-first-case-study-is-to-the-left-of-the-related-projects-title


jsms90:
LGTM

PR #85, opened 2019-01-18
feat: add CompactVideoLink

antoniocapelo:
## Description - [Trello ticket](https://trello.com/c/QKvyjeKI)

This PR aims to implement a reusable component called `CompactVideoLink` as documented here: https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5c3e05702c9f7852505e518f.

This component will be a dependency of a `StandaloneVideoLink` which will be used on the 'Join Us' page.

I created a storybook story for testing/checking the component in an isolated way.

This PR also fixes the following visual issue on the Talks component, it seems the different video links are not vertically aligned with each other.
image

(at least I think it's a bug @SaraVieira @jsms90 please shout if I'm wrong)

**Please note**: this MR adds 2 svgs for the play button (dark/light). I'm not comfortable with this, since the correct way to do this would be to have a single SVG for the icon, and change the `fill`s accordingly. However, I don't think we had the need to load `svg` files until now (every svg was being added as a `src` of an `img` tag). I think it's best to study the best svg loader solution, but that should be done on separate MR, in my opinion.

## Review template

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,7 +5,7 @@ module.exports = {
   moduleNameMapper: {
     '.+\\.(css|styl|less|sass|scss)$': 'identity-obj-proxy',
     '.+\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
-      '<rootDir>/.jest/__mocks__/fileMock.js'
+      '<rootDir>/.jest/__mocks__/fileMocks.js'
   },
   testPathIgnorePatterns: [
     '<rootDir>/node_modules',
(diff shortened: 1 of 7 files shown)


antoniocapelo:
@SaraVieira just did

jest.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   moduleNameMapper: {
     '.+\\.(css|styl|less|sass|scss)$': 'identity-obj-proxy',
     '.+\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
-      '<rootDir>/.jest/__mocks__/fileMock.js'
+      '<rootDir>/.jest/__mocks__/fileMocks.js'
```
antoniocapelo: this broke my head 😑

PR #89, opened 2019-01-21
Docs scripts

SaraVieira:
Closes https://trello.com/c/T5IXJZ4u/270-document-our-npm-scripts-how-they-should-be-used-whilst-developing

Also re-adds lighthouse and fixes accessibility issues


src/pages/index.js
```diff
@@ -27,47 +27,52 @@ const IndexPage = ({
...
-      <Padding bottom={{ smallPhone: 0, smallTablet: 2, desktop: 2 }}>
-        <CaseStudyPreview
-          caseStudy={content.featuredCaseStudy}
-          subHeading="Featured work"
-        />
-      </Padding>
-      <Padding bottom={{ smallPhone: 0, desktop: 2 }} />
-    </Grid>
-    <GreyBackground topMargin>
+    <main>
```
antoniocapelo: was this main added because of a lighthouse check? If so, shouldn't it be present on our `Layout` instead?

SaraVieira: Good point! Done

.circleci/config.yml
```diff
@@ -1,49 +1,54 @@
...
+    working_directory: ~/repo
+version: 2
 workflows:
-  version: 2
   build_and_test:
     jobs:
       - lint
+      - lighthouse
       - snapshots
+  version: 2
```
antoniocapelo: do we need this line and line 47 as well? since yaml is indent-based, I'm afraid this is redundant? Or probably I'm missing something :)

PR #95, opened 2019-01-23
Align typographic components with design docs

jsms90:
## Previous situation
**Designs**
[Zeplin](https://zpl.io/VD4g8Bg):
image

**Codebase**
- H1 - size corresponded to SectionTitle
- H2 - size corresponded to SectionTitle
- H2 with `small` prop - sizes didn't correspond to any particular Zeplin components (roughly used in place of DisplayTitle)
- H3 - size corresponded to DisplayTitle, but color was `textLight` (a grey) by default instead of `text` (i.e. black)
- H4 - size corresponded to CardTitle
- H5 - size corresponded to Subtitle
- H6 - size corresponded to BodyStylised, used in only one location (`SEOLinks`)
- Paragraph - corresponded to BodyPrimary

## Inconsistencies
**Codebase issues**
This ticket was originally raised by Tom because, although the sizes roughly correspond, the paddings were not consistent between Zeplin designs & the codebase.

But these "code component" to "design component" mappings were not always used consistently. Sometimes there were overrides to the typography's size at the page-level. Some added props don't seem to be necessary.

**Design issues**
It is not always clear from Zeplin, which typographic component engineers were supposed to use, as the sizes do not always directly correspond to one of the pre-defined components.

Paddings are also not consistent between different pages, which has led to engineers adding the `noMargin` prop to all typographic components.


jsms90:
Note: snapshot tests are failing because I haven't been updating them as I went. Since I will just need to override all the snapshots, it seemed sensible to do this at the end.

So, before I bother updating them, I just wanted to give @SaraVieira and designers a chance to feed back on whether this new component naming was ok.

@SaraVieira When I first floated the idea of renaming these components to match designers' naming, you made a good point that this would mean losing the semantics. So I've tried to address that by calling the components `DesignerName + Default semantic element` (e.g. `SubtitleH3`).

That way, it's really clear for an engineer who is using `Subtitle`, what heading level is under the hood by default.

In every situation where we need to use a different semantic value, I've used `as="hX"`.

Please do let me know what you think of this approach & whether it adequately addresses what you were telling me? :pray:

jsms90:
So, I notice that a lot of our "molcules" (taking the word from Atomic Design, for want of a term that we already use in this project) are being put together from Typography components that _don't_ have the paddings [defined in Zeplin](https://zpl.io/VD4g8Bg).

So it seems like we should be creating 2 separate types of typography component:
- with padding
- without

At the moment we achieve this like:
- `<H1>`
- `<H1 noMargin>`

But this runs the risk of engineers who are new to the project, forgetting the `noMargin` prop.

I think it would be a better idea to actually separate these into 2 components in the codebase, that are used in different situations.

Had a chat to Tom & Antonas about this and we came up with this naming convention
- `TypographyName` (e.g. `BodyPrimary`) - component with padding)
- `TypographyNamePrimitive` (e.g. `BodyPrimaryPrimitive`) - component without padding, to be used when building molecules, as well as inside the TypographyName components

antoniocapelo:
@jsms90 so we'd have `SubtitleH3Primitive` ?
I'm not 100% onboard with that naming. If I saw a `SubtitleH3Primitive` I'd think that it'd be like an inner component used by `SubtitleH3`..

jsms90:
That's exactly what it would be @antoniocapelo :+1: 

`SubtitleH3Primitive`
image

`SubtitleH3`
image

`SubtitleWithBody` = SubtitleH3Primitive + BodyPrimaryPrimitive
image

Does that make sense?

src/components/Typography/modifiers.js
```diff
@@ -0,0 +1,30 @@
+import { css } from 'styled-components'
```
antoniocapelo: 👍 for this file

antoniocapelo:
@jsms90 I understand the need for those 'no margin' exceptions and we should make it as easy as possible to apply them, but `SubtitleH3Primitive` as a name seems a bit confusing for engineers who are new to the project. 

I know that the `noMargin` can be overlooked (although that can be solved by documenting it on storybook) but for a 'React Developer' it's not clear what a `Primitive` component really is. There are some conventions of exporting a 'non-wrapped' componte as `PureX` where X is the name of the default wrapped component, but Primitive is not so clear. At least that's my opinion

src/components/Typography/index.js
```diff
@@ -1,7 +1,7 @@
-export { default as H1 } from './h1'
-export { default as H2 } from './h2'
-export { default as H3 } from './h3'
-export { default as H4 } from './h4'
-export { default as H5 } from './h5'
-export { default as H6 } from './h6'
-export { default as Paragraph } from './p'
+export { default as SectionTitleH1 } from './SectionTitleH1'
```
antoniocapelo: Might be my lack of food, but could a possible for solution for naming be the following:

- We only have one H1, so it should be `<H1>`
- We'd have one `<H2>` with a modifier for `display` (it would fallback for section)
- We'd have one `<H3>` with a modifier for `card` (it would fallback  for the subtitle)
- The body ones would stay the same.

I'm not so close to the designer so I might be missing something, if so, don't mind me!

src/components/Typography/modifiers.js
```diff
@@ -0,0 +1,30 @@
...
+  ${is('muted')`
+    opacity: .5;
+  `};
+
+  ${is('noPadding')`
+    padding: 0;
+  `};
+
+  ${is('noPaddingTop')`
+    padding: 0;
```
antoniocapelo: shouldn't this be `padding-top: 0`?

src/components/Typography/index.js
jsms90: I think one of the desired effects of this PR is to separate styles from heading hierarchy as far as possible.

We have to give each of these sets of styles a default heading number. So I've gone with the one that these styles are most commonly applied to. For example, we usually use the Subtitle styles on an HTML element that happens to be `h3` in the hierarchy.

For all other situations, we can use the `as="hX"` prop.

In our situation, we rarely actually use an h1 anywhere on the page, because this is reserved for SEO & is therefore only used on:
- the HidddenText element (which is not visible to a user and therefore needs no styles attached)
- the `SEOText` component

So, arguably, we don't actually need `SectionTitleH1` at all.

Does that makes sense? Happy to have a call about it tomorrow with all designers & engineers, to clarify the situation.

jsms90: resolved via a call between all engineers

jsms90:
In a call with @SaraVieira & @antoniocapelo last week, we discussed:
- Naming - remove semantics completely (e.g. H1) & just use `SectionTitle`, `CardTitle`, etc.  
There were some reservations about the existing names though - to be discussed together with Antonas
- Introducing 2 sets of typography components, one set with padding & one set without (Primitives) - we decided to keep the `noMargin` prop approach and not introduce another set of components. As we componentise the existing codebase, the situations in which we want a "Primitive" / noMargin version, will be encapsulated in the larger component

src/components/Typography/CardTitle.js
```diff
@@ -0,0 +1,38 @@
...
+    font-weight: 400;
+  `};
+
+  ${breakpoint('tablet')`
+    font-size: ${remcalc(26)};
+    line-height: ${remcalc(30)};
+    padding: ${remcalc(12)} 0 ${remcalc(12)} 0;
+    ${modifiers}
+
+    ${is('bigger')`
```
jsms90: Engineers, please note that this prop is only used in the "related projects" section on the SEO template & is likely to be an unintended design discrepancy.

Spoken to Antonas about this. Should get ironed out once he picks up [the trello ticket for that component](https://trello.com/c/9uQ3q3Fk/306-casestudycard-component).

So I've also added a note about (potentially) removing this prop, to the checklist on that trello ticket.

Hope that's clear enough?

src/components/Typography/CardTitle.js
```diff
@@ -0,0 +1,38 @@
...
+      font-size: ${remcalc(30)};
+      line-height: ${remcalc(36)};
+    `}
+
+    ${is('biggest')`
```
jsms90: Similar - this is only used in the Events section of the homepage & is likely to be down to design discrepancies.

So this prop will also probably be removed as part of [that trello ticket](https://trello.com/c/9uQ3q3Fk/306-casestudycard-component)

PR #101, opened 2019-01-25
Storybook deploy

SaraVieira:


SaraVieira:
Fuck now, it only works with npm

I like yarn

PR #109, opened 2019-01-30
review cats

SaraVieira:
## Description - [Trello ticket](https://trello.com/c/oSGSytSv)

Current Model is:

A training page has training categories and then each category has courses inside

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "*.js": [
       "prettier --write",
       "eslint --fix",
-      "jest --bail --findRelatedTests",
+      "jest --testPathPattern=\"storyshots.test.js\" --silent",
       "git add"
     ]
   }
--- a/src/components/Training/Courses.js
+++ b/src/components/Training/Courses.js
@@ -6,32 +6,31 @@ import Image from '../Common/Image'
 import StyledLink from '../styledLink'
 import { Grid, Row, Col } from '../grid'
 
-const Courses = ({ categories, courses, toggleModal }) => (
-  <Grid>
-    <Row>
-      <Col width={[1]}>
-        <Padding
-          bottom={{ smallPhone: 3, tablet: 4 }}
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -121,7 +121,7 @@
     "*.js": [
       "prettier --write",
       "eslint --fix",
-      "jest --bail --findRelatedTests",
```
SaraVieira: Jest never finds failing snapshots because they are all in one place so just run them everytime

src/components/Training/Courses.js
```diff
@@ -6,44 +6,43 @@ import Image from '../Common/Image'
...
+                <SectionTitle>{cat.name}</SectionTitle>
               </Padding>
               <Flex column alignStart>
-                {courseInCat.map((course, id) => (
+                {cat.courses.map(course => (
                   <StyledLink
                     style={{ cursor: 'pointer' }}
-                    onClick={() => toggleModal(course)}
-                    key={id}
+                    onClick={() => toggleModal({ ...course, ...cat })}
```
SaraVieira: Need the category logo and name in the modal

src/components/Training/Modal.js
```diff
@@ -79,8 +79,18 @@ const Close = styled.button`
   `}
 `
 
+const ModalStyles = createGlobalStyle`
```
SaraVieira: Cleaner way pf setting global styles

src/components/Training/Courses.js
antoniocapelo: of *course*

src/components/Training/Modal.js
antoniocapelo: this rules target the body? nice

SaraVieira: Global styles now allow for props! Its great!

PR #110, opened 2019-01-30
Had component

SaraVieira:
## Description - [Trello ticket](https://trello.com/c/eQGZS6da)

 Also adds better tests for the animated link and adds tests and stories for the case study card

Added jest styled components to get the actual styles in diffs and not classes


src/pages/contact.js
```diff
@@ -83,25 +83,12 @@ class ContactUs extends Component {
 
   render() {
     const { name, email, message, submitting, success } = this.state
-    const site = this.props.data.site
     const page = this.props.data.allContentfulPage.edges[0].node
     const { location } = this.props
```
antoniocapelo: might as well gather the data.site on this line, as well

src/pages/contact.js
```diff
@@ -83,25 +83,12 @@ class ContactUs extends Component {
...
     return (
       <Layout location={location}>
+        <Head site={this.props.data.site} page={page} />
```
antoniocapelo: why do we need to pass the site? isn't that done by the query on the `Head` ?

src/pages/contact.js
SaraVieira: Made it all easier by just getting one and not all

PR #117, opened 2019-02-04
add StatementComponent

SaraVieira:
## Description - [Trello ticket](https://trello.com/c/1q1UDBvN)

## Review template


src/components/Common/StatementComponent.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react'
```
antoniocapelo: Why the name `StatementComponent.js` and not just `Statement.js` ?

src/components/Common/StatementComponent.js
```diff
@@ -0,0 +1,57 @@
...
+      <Row>
+        <Col width={[1, 1, 1, 10 / 12, 10 / 12, 9 / 12]}>
+          <Padding
+            vertical={{
+              smallPhone: noPadding ? 0 : 3,
+              smallTablet: noPadding ? 0 : 4
+            }}
+          >
+            <DisplayTitle as="h1" textLight>
+              {typeof text === 'string'
```
antoniocapelo: just a sugestion. Wouldn't be better to use the `children` prop instead of the `text` one, when it's just a string?
So it'd become:

```
<DisplayTitle as="h1" textLight>
    { children }
    { text && text.map(....) }
</DisplayTitle>

```

antoniocapelo: and the usage is clearer (I guess)

```
<DisplayTitle>
  some text here
</DisplayTitle>

src/components/Common/StatementComponent.js
SaraVieira: idk, make sure everyone knows its a component ahah

PR #119, opened 2019-02-05
responsive fix in the menu

SaraVieira:
## Description - [Trello ticket](https://trello.com/c/q91rtsQF)

screenshot 2019-02-05 at 10 23 51

## Review template

### DIFF
--- a/src/components/Header/elements.js
+++ b/src/components/Header/elements.js
@@ -145,13 +145,15 @@ export const DesktopMenu = styled(Flex)`
   left: 0;
   top: 0;
   flex-direction: column;
-  padding: ${remcalc(24)};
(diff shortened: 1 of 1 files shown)


SaraVieira:
iPhone 5
screenshot 2019-02-05 at 10 31 56

antoniocapelo:
I like the code, but this is how it looks on my iPhone

image

This happens because of the non-sense bottom bar on safari mobile. 

I think the most correct solution would be to browser detect and compensate with the safari bottom bar height. 

**BUT** this is an overkill (we do not have browser detect lib nor are we needing it anywhere else on the site). So I suggest we do not center it vertically and just align it to the top. What do you think?

PR #130, opened 2019-02-08
Fix spacing & layout issues on case study template

jsms90:
## Spacing and layout issues on Engineering Case Study - [Trello ticket](https://trello.com/c/EYBREVnD/376-spacing-and-layout-issues-on-engineering-case-study)

To solve the problem:
- All text is now right aligned, instead of left aligned (and stats next to second text block is now on left hand side)
- Show/hide different versions of SecondTextSection based on screen size, now that Stats are on the left on desktop & on the bottom on mobile
- Simplify `Col` component (allow it to fall back to `display: flex` instead of applying `flex: 1`)
- Add correct spacing between different sections on mobile, using `Padding`
- Removed extra Padding/Margin elements

Refactor, while I was there:
- Separated CaseStudy template into smaller chunks for readability
- Improved semantics for the statistics - these used to be headings

---

## Review template


src/components/TemplatedCaseStudy/SecondTextSection.js
```diff
@@ -0,0 +1,75 @@
...
+const NonMobileRow = styled(Row)`
+  display: none;
+
+  ${breakpoint('smallTablet')`
+    display: flex;
+  `}
+`
+
+const MobileOnlyRow = styled(Row)`
+  display: flex;
```
antoniocapelo: I don't think this line is needed

src/components/TemplatedCaseStudy/SecondTextSection.js
```diff
@@ -0,0 +1,75 @@
...
+  <Col width={[1, 1, 1, 1, 9 / 12, 7 / 12]}>
+    {textBelowVideo.map((text, i) => (
+      <BodyPrimary key={i}>{text}</BodyPrimary>
+    ))}
+  </Col>
+)
+
+const SecondTextSection = ({ hasStats, stats, textBelowVideo }) => (
+  <Fragment>
+    {hasStats ? (
```
antoniocapelo: why do we need this bool? Isn't this achieved by checking if `!!stats`?

src/components/TemplatedCaseStudy/SecondTextSection.js
```diff
@@ -0,0 +1,75 @@
...
+      <Fragment>
+        <NonMobileRow spaced>
+          <Col width={[0, 0, 0, 0, 3 / 12]}>
```
antoniocapelo: I think we don't need the `0`s, just a single `[3/12]` is enough since the whole row will disappear on mobile? The same applies to line 62

jsms90: Yep true :+1:

I used to have this as a NonMobileCol & didn't think about it when I changed it to the entire row that's shown hidden.

Good stuff

src/components/TemplatedCaseStudy/FirstTextSection.js
```diff
@@ -0,0 +1,24 @@
...
+import { BodyPrimary } from '../Typography'
+import { Row, Col } from '../grid'
+
+const RowLayout = styled(Row)`
+  ${breakpoint('smallTablet')`
+    justify-content: flex-end;
+  `}
+`
+
+const FirstTextSection = ({ textAboveVideo }) => (
```
antoniocapelo: just a matter of style, I guess, but it seems `textAboveVideo` could be simply called `text` since this compo does not have the notion of any video (that's on its parent responsibility). No problem to keep it like this though :)

src/templates/caseStudy.js
```diff
@@ -1,72 +1,53 @@
...
-                    key={i}
-                    className="video-container"
-                    dangerouslySetInnerHTML={{
-                      __html: `<iframe width="844" height="480" src="${text}" frameBorder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowFullScreen></iframe>`
-                    }}
-                  />
-                ))}
-              </Col>
-            </Row>
+          <VideoSection videoInfo={videoInfo} />
```
antoniocapelo: 👍 for creating these subcompos

src/components/TemplatedCaseStudy/SecondTextSection.js
antoniocapelo: cool!

src/components/TemplatedCaseStudy/FirstTextSection.js
jsms90: Nah, I totally agree. Thanks :+1:

jsms90: Lemme know what you think of the changes

src/components/TemplatedCaseStudy/SecondTextSection.js
```diff
@@ -37,34 +37,34 @@ const Stats = ({ stats }) => (
...
+const SecondTextSection = ({ stats, text }) => (
   <Fragment>
     {stats ? (
       <Fragment>
         <NonMobileRow spaced>
           <Col width={[3 / 12]}>
             <Stats stats={stats} />
           </Col>
-          <TextColumn textBelowVideo={textBelowVideo} />
+          <TextColumn text={text} />
```
antoniocapelo: 💅

src/components/TemplatedCaseStudy/FirstTextSection.js
antoniocapelo: I like them, you got my blessing!

PR #134, opened 2019-02-08
fix: using react-grid-debugger-packages and small grid fixes

antoniocapelo:

## Review template


src/components/grid.js
```diff
@@ -108,27 +108,26 @@ const stylers = () => ({
 })
 
 const GridStyled = styled.div`
+  ${breakpoint('smallPhone')`
```
jsms90: Any reason for replacing line 131 with this?

What happens now, below the smallPhone breakpoint? (not that there are any phones that size, I assume)

jsms90: Oh I didn't realise smallPhone was 0 :laughing: 

Never mind :+1:

PR #135, opened 2019-02-11
Add/update design specialities

jsms90:
## Update design specialities & change 'technology used' to 'specialities involved' on CaseStudyHero - [Trello ticket](https://trello.com/c/Z9PobmKF/250-2-update-specialities-applied-aka-casestudyhero-aka-technologies-used)

- get specialities when querying Contentful for all design case studies (non-templated)
- show these new design specialities on CaseStudyHero
  - change subtitle from "technology used" to "specialities involved", as this subtitle now covers engineering specialities + design specialities
  - remove the piece of code that checks whether specialities have come from Contentful before displaying it (this is now a required field in the content model)
- alter the order of services & specialities on CaseStudyHero - services now come first, as this is the umbrella term above specialities
- alter the "We work with" section of the service template, to account for the fact that the design page now has only 1 column and no subtitles

Note: most of the work for this trello ticket were things to be done directly within Contentful

## Review template

### DIFF
--- a/src/components/Common/CaseStudyCards/CaseStudyHero.js
+++ b/src/components/Common/CaseStudyCards/CaseStudyHero.js
@@ -22,10 +22,8 @@ const SubSection = ({ heading, items }) => (
 
 const SpecialitiesAndServices = ({ specialities, services }) => (
   <Padding top={{ smallPhone: 0, tablet: 0.5 }}>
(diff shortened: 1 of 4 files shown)


jsms90:
Approval from Antonas :tada:

PR #138, opened 2019-02-18
fix: remove elements file and put each component on its own

antoniocapelo:
## Description - Our header component had some its internal components distributed amongst separate files and a global `elements.js` file, which caused some confusion. 

This MR aims to eliminate the `elements.js` file.

## Review template


SaraVieira:
Who was confused about this? Its a pretty normal pattern

antoniocapelo:
> Who was confused about this? Its a pretty normal pattern

ME! 😄 The reason for this changes is that before the last PR that was merged, we had
- a `MobileMenu` which was just an hamburger button
- a `DesktopMenu` which was both present on desktop and mobile
- a Close button which was only used on mobile but figured as it served for both resolutions
- `anchors` being styled directly by a generic CSS instead of with styled components

All of these were fixed on the last PR. While fixing those, new files were created and this `elements.js` file was just left with 2 components. 

I'm not against having an `elements.js` file, but I'm against having a `elements.js` which is 300 line long — in that case, I think it should be broken into pieces.

PR #139, opened 2019-02-18
add new about us page

jsms90:
Opening this PR, just to show the current status of the About Us page. The approach we are taking is to branch off of this, to create the sub-components & only merge this PR once the whole page is complete.

## Add About Us page - [Trello ticket](https://trello.com/c/oNCfRzBS/18-about-us-page)

- created `about-us.js` file - page only contains `Head` component for now
- queried Contentful for About Us content and passed into page, for use in sub-components


jsms90:
At the moment, when any photos are uploaded to Contentful that are not square, this is how it looks:
image

Just making a note here, for when we get around to documenting how content should be uploaded in epic 3.7

PR #140, opened 2019-02-18
Feat/topnav margins

antoniocapelo:
## Description - [Trello ticket](https://trello.com/c/2A7WSztO)

Instead of having a fixed margin, this approach is based on fixing a max width on our navbar items and let the content flow on the available space, this way they can stay a bit close to each other, if needed.

However, in order to not let the links be too close to each other, I'm scaling a bit the logo on the 'problematic' widths so that it stays a bit less crowded.

## Review template

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -20,6 +20,16 @@ const HomeLink = styled(HeaderAnchor)`
     display: none;
   `};
 `
+
+/* Note: This scale applied to the logo is a temporary solution until the topNav is rebuilt
+   according to new design specs. */
+const LogoLink = styled(Link)`
+  @media (min-width: 901px) and (max-width: 1005px) {
+    transform: scale(0.7, 0.7);
+    margin-left: -${remcalc(30)};
+  }
+`
+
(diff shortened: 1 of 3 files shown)


src/components/Header/index.js
```diff
@@ -67,6 +78,9 @@ const Header = ({ path, blue, logoColour }) => {
                   <HeaderAnchor activeClassName="active" to="/join-us/">
                     Join Us
                   </HeaderAnchor>
+                  <HeaderAnchor activeClassName="active" to="/join-us/">
+                    About Us
```
jsms90: @antoniocapelo I'm guessing that this was only in here for the purposes of development? i.e. checking what it looked like with an extra link?

We need to remove this, right? If the PR is going into `master`, rather than the `about-us-page` branch?

antoniocapelo: @jsms90 yes, I just put it there for antonas to see how it will behave in the future, which he approved this morning :-) 

image

So if you approve it, I can remove that extra link and merge it :-)

PR #141, opened 2019-02-18
Fix/topnav color on seo pages

antoniocapelo:
## Review template


antoniocapelo:
Thanks for the inpu and merget! Yeah, I agree we need a ticket/space on sprint to understand the best strategy to deal with this (theme object vs `themeVariation` vs other solution) and also see when's the best timing to tackle this.

Regarding the Hamburger button color, I also noticed that and went back to older commits and saw that it was like that from a long time. I bet that it was never right since we're using an `img` for the svg, which doesn't let us change the `fill` so it would always stay with the same color. 

I've also warned antonas regarding this, and will create a ticket for it.

PR #142, opened 2019-02-18
add subsidiaries section

jsms90:
## Subsidiary companies section - [Trello ticket](https://trello.com/c/u0IXjAlj/401-5-subsidiary-companies-section)

- fix graphql query - Contentful has nested objects for text fields :confused: 
- create SubsidiariesSection component in AboutUs folder using `ColumnLayout`
- only display 'Learn more' link if external link is provided in Contentful
- added logos to Contentful & display these within SubsidiariesSection

Refactor:
- Move BG components & StyledLink into `Common` folder

## Review template


src/components/AboutUs/SubsidiariesSection.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
```
antoniocapelo: I think this is the first time we give the suffix `Section` to a component which represents a section of a page, probably it's best to call the file just `Subsidiaries.js` to comply with the rest?

src/components/AboutUs/SubsidiariesSection.js
```diff
@@ -0,0 +1,66 @@
...
+          {({ Col, item: subsidiary }) => {
+            const {
+              description: { description },
+              image,
+              linkUrl,
+              linkText
+            } = subsidiary
+
+            return (
+              <PaddedCol block={false}>
```
antoniocapelo: Probably this whole block could be replaced by a component called `Subsidiary` ?

src/pages/about-us.js
```diff
@@ -3,12 +3,21 @@ import { StaticQuery, graphql } from 'gatsby'
...
-  </Layout>
-)
+const AboutUs = ({ data: { contentfulAboutUsPage: content } }) => {
+  const { yldGroupTitle, subsidiaries } = content
+
+  return (
+    <Layout>
+      <Head page={content} />
+      <SubsidiariesSection
+        yldGroupTitle={yldGroupTitle}
```
antoniocapelo: `yldGroupTitle` could be replaced by a simple `title` prob since it'll be used in the Subsidiaries component scope, so there's no need to create more complex prop names, what do you think?

jsms90: Haha same as last PR. I'll learn :wink: 

Thanks!

src/components/AboutUs/SubsidiariesSection.js
jsms90: Good point. I thought I had seen this pattern somewhere :thinking: 

Never mind

PR #144, opened 2019-02-19
Technology partnerships

jsms90:
## Technology partnerships page section - [Trello ticket](https://trello.com/c/ahG6KZbx/398-5-reuse-4-column-grid-logo-grid-for-our-partnerships)

- Created `Partners` sub component for About Us page
- Re-used `PaddedCol` pattern from Subsidiaries section, but didn't use `ColumnLayout` this time

FYI, looks like this if a 5th gets added

---

image

---

## Review template

### DIFF
--- a/src/components/AboutUs/Partners.js
+++ b/src/components/AboutUs/Partners.js
@@ -38,7 +38,11 @@ const Partners = ({ partnershipsTitle, partners }) => (
             block={false}
             width={[1 / 2, 1 / 2, 1 / 2, 1 / 2, 3 / 12]}
           >
-            <Image image={partner.image} width="250px" />
+            <Image
+              image={partner.image}
+              width="250px"
+              style={{ filter: 'grayscale(1)', saturate: '0' }}
+            />
           </PaddedCol>
         ))}
       </Row>
--- a/src/components/AboutUs/Partners.js
+++ b/src/components/AboutUs/Partners.js
@@ -1,29 +1,13 @@
 import React from 'react'
-import styled from 'styled-components'
-import breakpoint from 'styled-components-breakpoint'
 import { Padding } from 'styled-components-spacing'
 
 import Image from '../Common/Image'
-import { Grid, Col, Row } from '../grid'
+import { Grid, Row } from '../grid'
 import { SectionTitle } from '../Typography'
-
-const PaddedCol = styled(Col)`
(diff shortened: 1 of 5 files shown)


src/components/AboutUs/Partners.js
```diff
@@ -0,0 +1,53 @@
...
+    <Padding
+      top={{ smallPhone: 3, tablet: 4 }}
+      bottom={{ smallPhone: 3.5, tablet: 5 }}
+    >
+      <Padding bottom={{ smallPhone: 3, tablet: 4 }}>
+        <SectionTitle>{partnershipsTitle}</SectionTitle>
+      </Padding>
+      <Row>
+        {partners.map((partner, idx) => (
+          <PaddedCol
```
antoniocapelo: minor change, but I'd suggest doing the same as in the `Subsidiaries` — creating a `Partner` Component that does what this block does and use it here : )

src/pages/about-us.js
```diff
@@ -3,12 +3,18 @@ import { StaticQuery, graphql } from 'gatsby'
...
-    <Head page={content} />
-  </Layout>
-)
+const AboutUs = ({ data: { contentfulAboutUsPage: content } }) => {
+  const { partnershipsTitle, partners } = content
+
+  return (
+    <Layout>
+      <Head page={content} />
+      <Partners partnershipsTitle={partnershipsTitle} partners={partners} />
```
antoniocapelo: same as the other PR, `title` would be a cleaner prop name : )

src/components/AboutUs/Partners.js
jsms90: I get what you mean, but In the subsidiaries PR, the `PaddedCol` is not part of `Subsidiary`. If I follow the same pattern here, then `Partner` = `Image`

jsms90: Ok, I've just moved this into a separate sub-component called `Partner`

I want to copy this on Subsidiaries too - put `PaddedCol` inside Subsidiary, but for some reason that I cannot fathom, it breaks the layout :confused: 
image

jsms90: Ok, I see. It's this bit inside `ColumnLayout`
image

Probably, it's not worth using this component. I'm gonna replace it with an explicit `Row` + `Col`, like in Partners

PR #146, opened 2019-02-20
create AboutUsHero section & add necessary content fields to Contentful query

jsms90:
## AboutUsHero section - [Trello ticket](https://trello.com/c/DceUHQYf/394-5-hero-section)

- Added "statement text" & "supporting statement" blocks to Contentful query  
(Note: the number of "supporting statements" is hard-coded in Contentful - see trello ticket for product reqs)
- Created component for hero section

---

FYI, I didn't use the `Statement` component after all because it turns out there a fair few fundamental differences:

|  | `Statement` component | `AboutUsHero` |
| -- | -- | -- |
| Background colour | Grey | White |
| Text colour | Grey | Black |
| Typography | `DisplayTitle` | `SectionTitle` (slight error in Zeplin designs for size on mobile, but I checked with Antonas) |
| Column width | <li>small tablet = 10</li><li>largeTablet = 10</li><li>desktop = 9</li> | <li>small tablet = 8</li><li>largeTablet = 10</li><li>desktop = 8</li> |

## Review template

### DIFF
--- a/src/pages/about-us.js
+++ b/src/pages/about-us.js
@@ -25,17 +25,27 @@ const AboutUs = ({ data: { contentfulAboutUsPage: content } }) => {
     partners
   } = content
 
+  const supportingStatements = [
+    {
+      icon: supportingStatement1Icon,
+      text: supportingStatement1Text
+    },
+    {
+      icon: supportingStatement2Icon,
+      text: supportingStatement2Text
+    },
+    {
+      icon: supportingStatement3Icon,
+      text: supportingStatement3Text
+    }
+  ]
+
   return (
     <Layout>
       <Head page={content} />
       <AboutUsHero
         statementText={statementText}
-        supportingStatement1Icon={supportingStatement1Icon}
-        supportingStatement1Text={supportingStatement1Text}
-        supportingStatement2Icon={supportingStatement2Icon}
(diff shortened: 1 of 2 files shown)


src/pages/about-us.js
```diff
@@ -20,6 +28,15 @@ const AboutUs = ({ data: { contentfulAboutUsPage: content } }) => {
   return (
     <Layout>
       <Head page={content} />
+      <AboutUsHero
+        statementText={statementText}
+        supportingStatement1Icon={supportingStatement1Icon}
```
antoniocapelo: wondering if woud make sense to pass just an array of objects `{ icon, text}`, instead of passing six props?

jsms90: Yeah sure. I wondered about that. But then thought maybe it's not worth grouping these separate variables that come from contentful into an array & then taking them back out of the array the other side.

It is less code to change if/when the supporting statements become dynamic, so.... :balance_scale: 

Why not :+1:

jsms90: Mmmm, you're totally right. Now the `AboutUsHero` looks beautiful. Oh `map` :heart_eyes:

antoniocapelo: 💯

PR #147, opened 2019-02-20
Fix contact us page layout

jsms90:
## Incorrect layout for Contact Us page - [Trello ticket](https://trello.com/c/rZ2HD9P6/419-contact-us-page-layout-doesnt-quite-conform-to-designs-where-grey-background-starts)

- Moved title into its own section with white background - now using a prop to change the text of the title, instead of having one `SectionTitle` inside `Success` & one `SectionTitle` inside the form
- Deleted `Success` component & moved `BodyPrimary` (success message text) directly into render method
- Removed `Margin`s from around Grid & used same `Padding` pattern that we use on other pages
- Fixed spacing above & below grey section, for all breakpoints

Refactor along the way:
- Moved `forms.js` from `/components` into `/components/Common`
- Moved `Button` component out from `forms.js`, into its own file, since it is used on the 404 page too & could get used elsewhere in future (we don't have another Button component atm)
- Separate `TitleSection`, `ThankYouMessage` & `AreasOfInterest` into separate files

## Review template

### DIFF
--- a/src/components/Common/Button.js
+++ b/src/components/Common/Button.js
@@ -12,7 +12,7 @@ const Button = styled.button`
   line-height: ${remcalc(24)};
   box-sizing: border-box;
 
-  :disabled {
+  &:disabled {
     opacity: 0.5;
   }
 `
--- a/src/components/Common/Button.js
+++ b/src/components/Common/Button.js
@@ -1,0 +1,20 @@
+import styled from 'styled-components'
+import remcalc from 'remcalc'
+
+const Button = styled.button`
+  border: 0;
+  display: block;
+  padding: ${remcalc(18)} ${remcalc(24)};
+  color: ${props => props.theme.colors.white};
+  background: ${props => props.theme.colors.text};
+  font-weight: bold;
+  font-size: ${remcalc(18)};
+  line-height: ${remcalc(24)};
+  box-sizing: border-box;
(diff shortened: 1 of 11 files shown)


jsms90:
I'm tempted to alter this page so that the content is dynamically changeable via Contentful. I know the checkboxes, for example, are something that Greta has asked to be able to change in the past.

Not sure what you think, since is not currently a priority, but wouldn't take long & I'm here anyway :smiley:

src/components/Common/Button.js
```diff
@@ -0,0 +1,20 @@
...
+  display: block;
+  padding: ${remcalc(18)} ${remcalc(24)};
+  color: ${props => props.theme.colors.white};
+  background: ${props => props.theme.colors.text};
+  font-weight: bold;
+  font-size: ${remcalc(18)};
+  line-height: ${remcalc(24)};
+  box-sizing: border-box;
+
+  :disabled {
```
antoniocapelo: does this work? Shouldn't it be `&:disabled` ?

jsms90: No idea. I was just moving the code into a different file :wink:

antoniocapelo: just tried and it works 🤷‍♂️  leave it as is 😄

PR #148, opened 2019-02-21
feat: about us - client section

antoniocapelo:
## Description - [Trello ticket](https://trello.com/c/BDRisfCW)

Built client testimonials section

## Review template


src/components/AboutUs/ClientTestimonial.js
```diff
@@ -0,0 +1,69 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Padding } from 'styled-components-spacing'
+
+import { Row, Col, Grid } from '../grid'
+import { SectionTitle } from '../Typography'
+import GreyBackground from '../Common/GreyBackground'
+
+const Video = styled.iframe`
```
jsms90: Could we amalgamate this & the approach for the TemplatedCaseStudy, where we use a specially created `VideoSection` component:

https://github.com/yldio/yld.io/blob/875de550c7704cbbde32d21e37e0f9e78a7d1475/src/components/TemplatedCaseStudy/VideoSection.js#L5-L19

src/components/AboutUs/ClientTestimonial.js
```diff
@@ -0,0 +1,69 @@
...
+  height: 0;
+  padding-bottom: ${props => (props.height / props.width) * 100}%;
+`
+
+const CenteredRow = styled(Row)`
+  justify-content: center;
+`
+
+const ClientTestimonialVideo = ({ video }) => (
+  <Padding top={{ smallPhone: 3, tablet: 4 }}>
```
jsms90: This is _the_ most minor thing & there's no official defined pattern or anything anyway. But the way I've been interpreting the designs is that there seems to be a bit of a pattern of `72px` below all section titles. So just a little FYI, each of the other sub sections of this page, including your `Teams` section, has this as `Padding bottom=...` applied to the title.

That's maybe just a nice little pattern to repeat, in case this section ever gets bigger or more things get added in.

As I say, just a silly little FYI :laughing:

antoniocapelo: makes sense, done

src/components/AboutUs/ClientTestimonial.js
antoniocapelo: I don't think this is justifiable...

the layout (num of cols) is different, in that section they map over an array whereas here I don't. The only thing that's similiar is really the `iframe` code. 

Also, I followed the same pattern that's used on the `talks.js` file — which I like more.

jsms90: Hmm I see what you mean. Tbh I have no idea why this is mapping over anything. We only want to show one video on the case study pages:
image

and the reason I'm mentioning it is because it looks identical to what we have on this page.

So maybe the point is less "amalgamate approaches" & more of a request to remove `VideoSection`, use what you've done here instead and put it into Common. What do you think?

antoniocapelo: I don't that much time to do it rn, but if you want to do it, please do :-)

jsms90: Kk :+1:

src/components/Common/VideoSection.js
```diff
@@ -0,0 +1,47 @@
...
+  position: relative;
+  height: 0;
+  padding-bottom: ${props => (props.height / props.width) * 100}%;
+`
+
+const CenteredRow = styled(Row)`
+  justify-content: center;
+`
+
+const VideoSection = ({ videoLink }) => (
```
antoniocapelo: looking at it, `src` would be a better prop name I guess since we're not linking to anything

PR #149, opened 2019-02-21
Fix case study component paddings

jsms90:
## Descrepancies between CaseStudyHero paddings - [Trello ticket](https://trello.com/c/xJc03g1d/418-discrepancies-between-the-distance-from-hero-components-the-navbar)

- added paddings to `Row` on `CaseStudyLayout` - use `isTop` prop to check which paddings should be applied, based on where the component sits in the page
- make `CaseSudyHero` & `CaseStudyPreview` components "sections" in their own right - i.e. use `Grid` to wrap the Rows and Cols
- remove extra code from each page where these components are used

## Review template


src/components/Common/CaseStudyCards/CaseStudyLayout.js
```diff
@@ -15,6 +16,20 @@ const FlexCol = styled(Col)`
 `
 
 const RowLayout = styled(Row)`
+  ${breakpoint('smallPhone')`
+    padding-top: ${props =>
+      props.isTop ? props.theme.spacing[3] : props.theme.spacing[3.5]};
```
antoniocapelo: haven't dug deep into every usage, but couldn't this `isTop` prop be replaced by a `:firts-child{}` css rule?

jsms90: Good point...that would be nice :thinking: 

Here are some situations in which `isTop` applies:
https://github.com/yldio/yld.io/blob/483e907f14143730aba9f2974b023e53dd8979d6/src/templates/service.js#L38-L42

https://github.com/yldio/yld.io/blob/483e907f14143730aba9f2974b023e53dd8979d6/src/pages/case-study/canon-giving-stories-a-new-form.js#L110-L115

https://github.com/yldio/yld.io/blob/483e907f14143730aba9f2974b023e53dd8979d6/src/templates/caseStudy.js#L26-L32

I'm guessing that, even though it's at the "top" of the page & the first child that's visible, that the `<Head>` component will be the actual first child?

src/components/Common/CaseStudyCards/CaseStudyLayout.js
```diff
@@ -15,6 +16,20 @@ const FlexCol = styled(Col)`
...
+    padding-bottom: ${props =>
+      props.isTop ? props.theme.spacing[3] : props.theme.spacing[3.5]};
+  `}
+
+  ${breakpoint('tablet')`
+    padding-top: ${props =>
+      props.isTop ? props.theme.spacing[3.5] : props.theme.spacing[5]};
+    padding-bottom: ${props =>
+      props.isTop ? props.theme.spacing[5] : props.theme.spacing[5]};
```
antoniocapelo: do we need this ternary?

src/components/Common/CaseStudyCards/CaseStudyLayout.js
jsms90: Just for record, Capelo pointed out that Head doesn't render anything. It’s just a call to the `Helmet` comp which update our `head` tag.

But we're sticking with this prop because the `:first-child` CSS solution relies on us building our pages in a particular way. For example, if a `CaseStudyPreview` component was used at the bottom of the page & was wrapped inside a div (e.g. `GreyBackground`), then it would be the first child of that div, but not the first child of the page.

src/components/Common/CaseStudyCards/CaseStudyLayout.js
jsms90: Oops. Good point. Just pushed the correct top & bottom paddings for "top" CaseStudy cards. Gonna look over this with Antonas when he is ready to review, because these values don't exist in `spacing` in our theme. So gotta check

jsms90:
Got the ok from Bea, in terms of design QA - https://trello.com/c/xJc03g1d/418-discrepancies-between-the-distance-from-hero-components-the-navbar :+1:

PR #150, opened 2019-02-22
bring CustomisedBullepoint in line with design specs, use this on appropriate pages & remove from other pages

jsms90:
## Review customised bullet point - [Trello ticket](https://trello.com/c/NUU4GCQk/319-3-review-customisedbulletpoint-component)

First attempt:
- move CustomisedBullepoint into Common
- use on training page instead of `BodyPrimary` + `Hr` + `Padding` with slightly different values from CustomisedBullepoint

After design QA:
- Fix `1px` discrepancy found in QA - this has been done by using our actual `Hr` component instead of an `:after` pseudo class
- Removed all the variations on `CustomisedBulletpoint` (spaced, symmetrical, etc) & updated stories accordingly
- Removed `CustomisedBulletpoint` from all situations where a list item with an `<hr>` is called for, but where the spec is NOT the same as the CustomisedBulletpoint spec

## Review template


src/components/Homepage/events/DesktopMeetups.js
```diff
@@ -18,19 +19,15 @@ const DesktopMeetups = ({ events }) => (
     <Padding top={42}>
       <ul>
         {getHomepageMeetups(events).map(event => (
-          <CustomisedBulletpoint
-            fullWidth
-            fullWidthBorder
-            symmetrical
-            key={`${event.id}`}
-          >
+          <li key={`${event.id}`}>
```
antoniocapelo: since this is also used on the homepage, would it make sense to create a component for this?

antoniocapelo: If it's not 100% the same, forget what I just said

stories/customisedBulletpoint.stories.js
```diff
@@ -7,23 +7,21 @@ addDecorator(Theme)
...
 storiesOf('CustomisedBulletpoint', module)
   .add('CustomisedBulletpoint', () => (
-    <CustomisedBulletpoint>I am a bullet thingy</CustomisedBulletpoint>
+    <div>
+      <CustomisedBulletpoint>I am a bullet thingy</CustomisedBulletpoint>
+    </div>
   ))
-  .add('CustomisedBulletpoint spaced', () => (
-    <CustomisedBulletpoint spaced>I am a bullet thingy</CustomisedBulletpoint>
+  .add('CustomisedBulletpoint with non-deafault width (500px)', () => (
```
antoniocapelo: typo :)

PR #152, opened 2019-02-27
Separate mobile and desktop navbar

jsms90:
## break nav component into mobile and desktop - [Trello ticket](https://trello.com/c/ktSJpL9c/426-5-break-topnav-component-into-mobile-and-desktop)

Branching off this PR to do other tickets in this epic

## Review template


src/components/Header/Hamburger.js
```diff
@@ -13,22 +13,24 @@ const HamburgerSvg = styled.svg`
...
 
-  ${breakpoint('tablet')`
+    &:focus {
+      background: transparent;
+      outline: ${remcalc(4)} solid ${props => props.theme.colors.outline};
+      color: ${props => props.theme.colors.text};
+    }
+  `}
+
+  @media screen and (min-width: 960px) {
```
antoniocapelo: now that we got our hands at it, we could try to use our `tablet` breakpoint (`901px`) instead of this 960... this way we could refer to it here instead of these 960px — which does not fit inside our breakpoint system. What do you think?

src/components/Header/Overlay.js
```diff
@@ -4,7 +4,7 @@ import is from 'styled-is'
 const Overlay = styled.div`
   display: none;
 
-  @media screen and (min-width: 600px) and (max-width: 900px) {
+  @media screen and (max-width: 959px) {
```
antoniocapelo: same comment as the previous here :)

src/components/Header/SideNavbar.js
```diff
@@ -0,0 +1,108 @@
...
+
+  &:not(:last-child) {
+    margin-bottom: ${remcalc(18)};
+  }
+
+  &:last-child > a {
+    padding-right: 0;
+  }
+`
+const SideNavModal = styled(Flex).attrs({
```
antoniocapelo: why the `Modal` name?

src/components/Header/index.js
```diff
@@ -37,79 +60,28 @@ const Header = ({ path, blue, logoColour }) => {
...
+            <FixedHeightFlex alignCenter justifyBetween as="header">
+              <Link to="/">
+                <Logo path={path} blue={blue} logoColour={logoColour} />
+              </Link>
+              <TopNavbar links={links} dark={blue} />
+              <Hamburger
+                light={!!blue}
+                onClick={() => toggleMobileNavbar(!isMobileNavbarOpen)}
+              />
+              <Overlay
```
antoniocapelo: Probably the `Overlay` and `SideNavbar` could be encapsulated by a `MobileNavigation` or `MobileMenu` component?

src/components/Header/Hamburger.js
jsms90: It's actually intentionally `960px` in the design specs, rather than 'tablet' because this is the width taken up by that many links + their padding.

I'm rebasing onto master right now & the way I'm fixing it (temporarily) is by making this `1010px` instead of `960px` because it's only going to stay like this for as long as we have such a larger number of links on the TopNav.

As soon as we pick up [the new navigation structure](https://trello.com/c/4O2nPwzt/388-3-navigation-structure-on-topnav-desktop) (i.e. add the dropdowns into the top nav), this can go back to `960px`. So I'm doing that ticket next.

src/components/Header/SideNavbar.js
jsms90: It behaves like a modal, no? :woman_shrugging: 

This `Flex` is the one with the background colour - it's the modal that pops up over the top of the rest of the page.

src/components/Header/index.js
jsms90: Yeah, it's a good point. Tbh I was really unsure about what to include in the definition of each `nav` anyway.

Is the TopNavbar just the set of similar-looking links? Or would you say the YLD logo is in the TopNavbar? Since there is a separate 'Home' link in the SideNav, perhaps the home navigation (YLD logo) _should_ be part of TopNav? Then what's the distinction between the `header` & `nav` in our case?

Would you say the hamburger is part of the TopNavbar or the SideNavbar? If you say that the `nav` is only the set of links, then is the blue/black background (what I'm calling the Modal) actually part of the SideNav?

Or maybe the SideNav is only the set of links, which sit inside the modal. In which case, the Overlay is no more part of the SideNav than the Modal is.

Not sure if that's particularly easy to understand, written down... but what do you think "nav" should mean?

src/components/Header/Hamburger.js
antoniocapelo: 👍

src/components/Header/SideNavbar.js
antoniocapelo: are you talking about this dark menu?

image

If so, I guess it should be called `SideNavPanel` or `SideNav`

src/components/Header/index.js
antoniocapelo: In a general approach, I use these rules of thumb:
- `nav` is literally a `nav` tag with a `ul` and `li`s with anchors.
- a component that has one of those `nav` inside, should be called `Navigation` (or something similar`

So, for this scenario, I'd imagine something like:
- `Header` is the top-level component that has
  - `MobileNavigation`
  - `DesktopNavigation`
  - `Hamburger`
  - `Logo(s)`

What do you think?

jsms90: Yeah I think we're on the same page really. This is what I've gone with:
> `nav` is literally a `nav` tag with a `ul` and `li`s with anchors

which is why the Hamburger is outside the TopNavbar & also why the Overlay is outside the SideNavbar.

Having done a bunch of googling before getting started, it feels like there's little to no convention out there & that everyone does whatever makes sense for their project. So I reckon it's all personal preference really. I'm happy to change it from this formulation to whatever you suggest.

Maybe it's easier to chat about this in person, when you get to London today?

src/components/Header/SideNavbar.js
jsms90: Yeah but the name of the wrapper `div` (SideNavModal/SideNavPanel) + its contents (close button + list of links) is already called `SideNav.

But if "Panel" is more accurate than "Modal" for this type of thing, I can change it to that :+1:

jsms90:
This PR is failing because of a single lighthouse test
image

This is about the dom size & our rule is that it shouldn't be larger than `500`. I'm not sure how these rules were decided upon and therefore whether this is something that we want to stick to? But having had a bit of a google, the general advice that I'm finding is that [over 1500](https://developers.google.com/web/tools/lighthouse/audits/dom-size) is when it's deemed "too large". So, one potential option might be to relax this rule?

Alternatively, one easy way of reducing our dom size would be to stop using [`styled-components-spacing`](https://github.com/jameslnewell/styled-components-spacing) which is a library that gives us use of `<Padding>` & `<Margin>` components. Under the hood, these are just `div`s that we wrap other components with, in order to create spacing. Instead, we could just add padding/margin to the relevant components with CSS.

A third option may be that I have made this component in a more complex way than is necessary?

@antoniocapelo any thoughts?

jsms90:
On an in-person call, we agreed to close this PRs in favour of a single PR encompassing all dropdown-related navbar tickets. See trello - https://trello.com/c/MLVz9RqD/432-13-implement-new-navbar-structure-including-dropdowns

On a separate note, in terms of [resolving the lighthouse test failures](https://github.com/yldio/yld.io/pull/152#issuecomment-468628931), we decided to go with option 1 - relaxing the rule to allow for more than 500 dom nodes.

PR #157, opened 2019-03-11
New nav structure

jsms90:
## Implement new navbar structure, including dropdowns - [Trello ticket](https://trello.com/c/MLVz9RqD/432-13-implement-new-navbar-structure-including-dropdowns)

Navbar:
- Split Header into `TopNav` & `SideNav` & pass in an object of links as a prop to both
- Split `DesktopDropdown` into 2 versions
- Implement new/updated spacings for TopNav (e.g. padding above and below logo & anchors)
- Implement new/updated designs for SideNav - colour, font size, spacing - everything
- Update styling for newly designed states in Zeplin. These styles live in a file called `outerItemStyles.js`, for re-use across anchors & the dropdown names
- Use `navItemSpacing.js` files to share spacing values between outer & inner list items - done via a combination of padding and margin, in order to make room for the outline styles on focus
- Use `960px` again, as the breakpoint width for `TopNav` turning into `SideNav`

Refactor, during build:
- Remove `outline` from theme & use `vibrant` everywhere in codebase. Also update the `vibrant` colour to "Neon green" from Zeplin
- Stop passing `activeClassName` prop through to `ExternalLink` (fixes console log error)
- Change the background colour of the sections on the Our Work page - hero section is now white & main section is grey
- Increase values of time to FCP & FMP on lighthouse tests - [performance issues to be investigated](https://trello.com/c/WazHvmYQ/446-investigate-lighthouse-failures)

## Review template


src/components/Common/ButtonAsIconWrapper.js
```diff
@@ -0,0 +1,20 @@
+import styled from 'styled-components'
+import remcalc from 'remcalc'
+
+const ButtonAsIconWrapper = styled.button.attrs({
+  type: 'button'
+})`
+  min-width: ${props => remcalc(props.theme.elementSizes.tappableArea)};
```
antoniocapelo: I'd move these two rules into our `UnstyledButton` and just add the centering directly  in the place where this component was being used

src/components/Header/Hamburger.js
```diff
@@ -1,40 +1,43 @@
...
-const HamburgerSvg = styled.svg`
-  fill: currentColor;
+import ButtonAsIconWrapper from '../Common/ButtonAsIconWrapper'
+import outlineStyles from './outlineStyles'
 
-  ${is('light')`
-    fill: ${props => props.theme.colors.white};
-  `}
+const HamburgerSvg = styled.svg`
+  ${props =>
```
antoniocapelo: couldn't this be simplified? 
```js
fill: ${ ({themeVariation, theme}) => themeVariation === theme.variations.dark ? theme.colors.white : theme.colors.text };
```

src/components/Header/Logo.js
```diff
@@ -19,13 +25,25 @@ const Logo = ({ path, logoColour, blue }) => {
       !path.includes('design') &&
       !path.includes('training') &&
       !blue ? (
-        <img role="link" tab-index="0" height="48" src={logo} alt="yld logo" />
+        <YLDLogoImageWrapper>
+          <img
+            role="link"
+            tab-index="0"
```
antoniocapelo: Do we need this tab-index? (also, react uses `tabIndex`)

src/components/Header/SideNav/CloseButton.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import styled from 'styled-components'
+import remcalc from 'remcalc'
+
+import close from '../../../images/close.svg'
+import ButtonAsIconWrapper from '../../Common/ButtonAsIconWrapper'
+import outlineStyles from '../outlineStyles'
+
+const CloseIcon = styled.img`
+  fill: ${props => props.theme.colors.white};
```
antoniocapelo: I don't think `fill` will work here since this is an img (it would if we were using an `svg` tag)

src/components/Header/SideNav/outerItemStyles.js
```diff
@@ -0,0 +1,18 @@
+import { css } from 'styled-components'
```
antoniocapelo: This file still seems strange being called `outerItemStyles.js` while it's being used by `OuterAnchorItem`, `Dropdown` and `InnerAnchorItem`... Can it be called `sidenavItemStyles` or smthing like that?

src/components/Header/TopNav/outerItemStyles.js
```diff
@@ -0,0 +1,68 @@
+import { css } from 'styled-components'
```
antoniocapelo: Same comment as on the other `outerItemStyles.js` file

src/components/Header/SideNav/index.js
```diff
@@ -0,0 +1,78 @@
...
+import Flex from 'styled-flex-component'
+import is from 'styled-is'
+
+import CloseButton from './CloseButton'
+import Dropdown from './Dropdown'
+import OuterAnchorItem from './OuterAnchorItem'
+
+const SideNavList = styled.ul``
+
+const SideNavPanel = styled.div.attrs({
```
antoniocapelo: why not `styled.nav`?

src/components/Header/SideNav/index.js
```diff
@@ -0,0 +1,78 @@
...
+import styled from 'styled-components'
+import remcalc from 'remcalc'
+import Flex from 'styled-flex-component'
+import is from 'styled-is'
+
+import CloseButton from './CloseButton'
+import Dropdown from './Dropdown'
+import OuterAnchorItem from './OuterAnchorItem'
+
+const SideNavList = styled.ul``
```
antoniocapelo: unnecessary?

src/components/Header/SideNav/index.js
```diff
@@ -0,0 +1,78 @@
...
+  z-index: ${props => props.theme.zIndexes.header};
+  transform: translateX(100%);
+  transition: transform ${props => props.theme.animations.fast} ease-in-out;
+
+  /* Thinner version' of the navbar */
+  @media screen and (min-width: 600px) and (max-width: 959px) {
+    width: ${remcalc(295)};
+    left: auto;
+    right: 0;
+    z-index: 10;
```
antoniocapelo: why do we need to change the `z-index` here?

src/components/Header/SideNav/index.js
```diff
@@ -0,0 +1,78 @@
...
+  <SideNavPanel open={isOpen}>
+    <Flex justifyEnd>
+      <CloseButton onClick={onClose} />
+    </Flex>
+    <SideNavList>
+      <OuterAnchorItem activeClassName="active" to="/">
+        Home
+      </OuterAnchorItem>
+      {links.map((link, idx) => {
+        if (link.dropdownItems) {
```
antoniocapelo: I'd join lines 53-71 into a `<SideNavItem item={link}>` that does the logic to check for dropdownItems inside

src/components/Header/TopNav/Dropdown.js
```diff
@@ -0,0 +1,190 @@
...
+  hoverDarkStyles,
+  clickTapDarkStyles,
+  activeAndHoverDarkStyles
+} from './outerItemStyles'
+
+const DropdownContainer = styled.li`
+  position: relative;
+  cursor: pointer;
+  background: transparent;
+  ${props => {
```
antoniocapelo: when looking at this, I got an idea: we could use [the attr method from styled comps](https://www.styled-components.com/docs/api#attrs), something like

```
const lightStates = {
default: defaultLightStyles,
click: clickTapLightStyles,
hover: hoverLightStyles,
expandedAndHover: activeAndHoverLightStyles
}

const darkStates = {....}

// (We could even export these in a single object, if possible)

const DropdownContainer = styled.li.attrs({
  states: props => (props.themeVariation === 'light' ? lightStates : darkStates),
})`
> span {
  ${props => props.states.default}
}
&:hover > span {
  ${props => props.states.hover}
}
....
`
```

What do you think?

src/components/Header/TopNav/OuterAnchorItem.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+`
+
+const StyledListItem = styled.li`
+  display: flex;
+  &:focus {
+    ${outlineStyles}
+  }
+
+  ${props => {
```
antoniocapelo: probably here we could the same as I suggested in Dropdown.js above

src/components/Header/TopNav/outerItemStyles.js
```diff
@@ -0,0 +1,68 @@
...
+import remcalc from 'remcalc'
```
antoniocapelo: so many styles 🤯

src/components/Header/Logo.js
jsms90: Oh yeah. Weird.

I didn't change this code though - the `` line is the same as the old line 22. I just wrapped it with YLDLogoImageWrapper.

Not sure why this was there in the first place :woman_shrugging: But you're right. I'll remove it.

src/components/Header/SideNav/CloseButton.js
jsms90: :woman_facepalming:

src/components/Header/SideNav/outerItemStyles.js
jsms90: Good point. `fontSizeAndWeight` doesn't really belong in this file, as it's used on "inner" items as well.

I've moved `fontSizeAndWeight` into the same file that contains `itemSpacing` (margin + padding) & renamed that file `sideNavItemStyles.js`.

See what you think

src/components/Header/SideNav/index.js
jsms90: Yeah, nice :+1: I just copy/pasted [the code from the old file](https://github.com/yldio/yld.io/pull/157/files#diff-3af6432a615e0c766f3ca19b44ffa57fL85). Lazy lazy :sleeping:

src/components/Header/SideNav/outerItemStyles.js
antoniocapelo: have you commited these changes yet?

src/components/Header/SideNav/index.js
jsms90: Oops. I think I only changed it when I was playing around, trying to fix the dropdowns not appearing in front of the page content.

[It used to be the `header` value from the z-index object in our theme.](https://github.com/yldio/yld.io/pull/157/files#diff-3af6432a615e0c766f3ca19b44ffa57fL95) But I don't know why a z-index needed to be applied in the first place.

src/components/Header/TopNav/outerItemStyles.js
jsms90: You talking about us having the `remcalc` library? Or just that we have so many styles files? :laughing: 
:boom:

src/components/Header/SideNav/outerItemStyles.js
jsms90: Oops. Hadn't pushed

src/components/Header/SideNav/index.js
jsms90: Ah no, [just copy/pasted](https://github.com/yldio/yld.io/pull/157/files/9764ac16c04781d47310dabd99aa86c7c6e70519#diff-3af6432a615e0c766f3ca19b44ffa57fL108). I'll remove it

src/components/Header/TopNav/outerItemStyles.js
antoniocapelo: so many js files 😄 I think that if it were repeated css classes, we'd not see it so badly but as we see so many files being exported/imported, it feels a bit scary haha

src/components/Header/SideNav/Dropdown.js
```diff
@@ -5,20 +5,16 @@ import is from 'styled-is'
 import Chevron from '../../Common/Chevron'
 import InnerAnchorItem from './InnerAnchorItem'
 import headerItemStyles from '../headerItemStyles'
-import sideNavItemSpacing from './sideNavItemSpacing'
-import {
-  fontSizeAndWeight,
-  defaultStyles,
-  hoverActiveStyles
-} from './outerItemStyles'
+import { fontSizeAndWeight, itemSpacing } from './sideNavItemStyles'
```
antoniocapelo: not to nag, but since we only export two objects on the `sideNavItemStyles`, and we use them together everytime, it would simplify to just export `default` a single object with all the rules and then use it everywhere?

antoniocapelo: If this logic could be applied on any other file, even better..!

src/components/Header/TopNav/Dropdown.js
```diff
@@ -95,7 +93,7 @@ const DropdownNameWrapper = styled.span`
     css`
       &:focus {
         outline-color: ${props.theme.colors.white};
-        ${topNavItemSpacing}
+        ${itemSpacing}
```
antoniocapelo: do we need `itemSpacing` here? isn't that set on line 87?

src/components/Header/TopNav/outerItemStyles.js
jsms90: Yep. It's true. I feel like it's just a weird, relatively unusual quirk of needing to re-use styles across different components - e.g. DropdownName and OuterListItem.

src/components/Header/SideNav/index.js
jsms90: Not sure if this is quite what you envisaged?

src/components/Header/SideNav/Dropdown.js
jsms90: Oh yeah good point. Something like `const listItemStyles` that encapsulates both :+1: 

Well, if you don't like the idea of having files that are purely for common styling, we could pick a component to "own" these styles. Say, we could:
1. put this `listItemStyles` inside `OuterAnchorItem.js` & export from there
2. delete the extra `listItemStyles.js` file
3. `import { listItemStyles } from 'OuterAnchorItem'` within InnerAnchorItem & Dropdown.js

If it's just a matter of not having more files than the number of actual components, then we could do that?

src/components/Header/TopNav/outerItemStyles.js
antoniocapelo: yeah, this is an unusual case. But I'm thinking to myself that if we were using "traditional" css,  we'd just apply some common classes (that would represent these common appearances) to these elements.. and that's essentially what we're doing by passing these object styles (with a bit overload of more css being outputed, I guess)

src/components/Header/TopNav/outerItemStates.js
```diff
@@ -1,61 +1,85 @@
...
+
+export const darkStates = {
+  default: defaultDark,
+  hover: hoverDark,
+  clickTap: clickTapDark,
+  active: activeDark,
+  activeAndHover: activeAndHoverDark
+}
+
+// const outerItemStates = themeVariation => {
```
antoniocapelo: remove?

src/components/Header/TopNav/outerItemStates.js
```diff
@@ -1,61 +1,85 @@
...
   background: ${props => props.theme.colors.white};
   color: ${props => props.theme.colors.text};
   opacity: 1;
 `
 
-export const hoverLightStyles = css`
+const hoverLight = css`
   background: ${props => props.theme.colors.greyBG};
   color: ${props => props.theme.colors.text};
   opacity: 1;
```
antoniocapelo: It seems that since the `opacity: 1` is on the default state, some of these `opacity: 1` rule can be remove, or am I wrong?

src/components/Header/TopNav/OuterAnchorItem.js
```diff
@@ -27,60 +16,35 @@ const StyledAnchor = styled(Anchor)`
...
   display: flex;
   &:focus {
     ${outlineStyles}
   }
 
-  ${props => {
-    if (props.themeVariation === 'light') {
-      return css`
-        ${defaultLightStyles}
+  ${props => props.states.default}
```
antoniocapelo: this turned out pretty elegant 💅

src/components/Header/TopNav/Dropdown.js
```diff
@@ -8,72 +8,32 @@ import InnerAnchorItem from './InnerAnchorItem'
...
-  hoverDarkStyles,
-  clickTapDarkStyles,
-  activeAndHoverDarkStyles
-} from './outerItemStyles'
-
-const DropdownContainer = styled.li`
+import { lightStates, darkStates } from './outerItemStates'
+
+const DropdownContainer = styled.li.attrs(props => ({
+  states: props.themeVariation === 'dark' ? darkStates : lightStates
```
antoniocapelo: 🚀

PR #161, opened 2019-03-13
training modals - fix column layout & remove width restriction on CustomisedBulletpoint

jsms90:
## training modals - fix column layout & remove width restriction on CustomisedBulletpoint - [Trello ticket](https://trello.com/c/UJ6SILfT/423-1-training-modals-fix-column-layout)

- remove width restriction from `CustomisedBulletpoint` on training page `Modal`
- instead restrict the width of the bullet points (and/or any other text) via the columns - i.e. bring Modal's column layout in line with design specs

## Review template

### DIFF
--- a/src/components/Training/Modal.js
+++ b/src/components/Training/Modal.js
@@ -95,9 +95,9 @@ const ModalStyles = createGlobalStyle`
 `
 
 const CourseInfo = ({ content }) => (
-  <Col width={[1, 1, 1, 1, 1 / 2]}>
-    <Image image={content.logo} width="60px" />
+  <Col width={[1, 1, 1, 1, 5 / 6]} px={0}>
     <Padding bottom={1}>
+      <Image image={content.logo} width="60px" />
       <SectionTitle>{content.name}</SectionTitle>
     </Padding>
     {content.description && (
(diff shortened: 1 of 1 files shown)


src/components/Training/Modal.js
```diff
@@ -153,8 +153,12 @@ const Modal = ({ content, toggleModal }) => (
         </Close>
         <Grid>
           <Row>
-            <CourseInfo content={content} />
-            <CourseContent content={content} />
+            <Col width={[1, 1, 1, 1, 1 / 2]}>
+              <CourseInfo content={content} />
```
antoniocapelo: with these lines 156-161, won't we have columns inside columns? is this what we want?

jsms90: I'm all ears, if you have other suggestions?

I see what you mean, with it seeming strange at first glacen. But the `Col` component is just a div underneath, so I guess there's no real problem there. And I figure this captures the intent nicely enough.

You can either think of it like:
1. split the page in half (6 columns and 6 columns)
2. allow content inside that half of the page span 5 columns

I reckon that version is more like how a designer thinks?

Or you can think of it like:
- text spans 5 columns
- 1 column of space
- text spans 5 columns
- 1 column of space

If we opt for this, then I guess we would have to create `div`s with nothing in them, to represent the column of space.

jsms90:
Reviewed & okayed by Bea :+1:

PR #166, opened 2019-03-22
give training course modals their own urls

GBouffard:
## Description - [8 - Training page modals need custom URLs](https://trello.com/c/9JPB74JF/440-8-training-page-modals-need-custom-urls)

Each training course have now their own urls.
- A course can be accessed via the training page and will open a modal with its own url (example: url can be http://localhost:8000/training/node-js/node-js-fundamentals)
- a course can be accessed via its url and will open as a page (example: accessed via http://localhost:8000/training/node-js/node-js-fundamentals).
In both cases, clicking the close button redirects to the training page of the site.

As part of the ticket, we also refactored: `src/templates/trainingCourseModal.js` which was a 200 lines legacy code component. The subcomponents have been moved to their own respective sub-files/sub-components:
- CourseWrapper
- CourseInfo (left column of the training course)
- CourseContent (right column of the training course)
- CourseCloseButton

## Review template


src/components/Training/course/CourseInfo.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { Padding } from 'styled-components-spacing'
+
+import StyledLink from '../../../components/Common/StyledLink'
+import Image from '../../../components/Common/Image'
+import { SectionTitle, BodyPrimary } from '../../../components/Typography'
+import { Col } from '../../../components/grid'
+import SubtitleWithBody from '../../../components/Common/SubtitleWithBody'
+
+// TODO - retrieve from contentful
```
antoniocapelo: is this TODO to be addressed on this PR?

antoniocapelo: either way, `constants` is not a very good name for this :)

src/pages/training.js
```diff
@@ -1,52 +1,26 @@
...
 
 const TrainingPage = ({ data: { contentfulTrainingPage: content } }) => {
-  const [modalContent, setModalContent] = useState(null)
   const trainingApproachesContent = [
     content.trainingApproachContent1.trainingApproachContent1,
     content.trainingApproachContent2.trainingApproachContent2,
     content.trainingApproachContent3.trainingApproachContent3
   ]
 
-  useEffect(
```
antoniocapelo: any reason for not having ported this to `trainingCourseModal`? We lost the functionality to hitting 'escape' to close the modal, didn't we?

src/pages/training.js
```diff
@@ -105,6 +77,7 @@ export const query = graphql`
         }
         courses {
```
antoniocapelo: the graphql query for the `courses` field just needs to include the `name`, `id` and `slug` properties, right? The rest are already fetched by the course page query

antoniocapelo:
One thing I noticed while playing with the preview environment is that when we click on a course to navigate to its page and then close the modal to navigate back, we're taken to the top of the page (and not where we were). 

A quick (but not ideal) solution would be to add an ID to the courses section and instead of navigating to /training, navigate to /training#courses (however this is not ideal and does not scale well)

src/components/Training/course/CourseInfo.js
GBouffard: @antoniocapelo I am not sure it need to be addressed in this PR. I noticed it was hardcoded and it would make sense to control such strings from contentful. We will think of another name; obviously, we are open to your suggestions for names :)

src/pages/training.js
GBouffard: We can discuss this in the stand-up as we need to discuss the implementation following our changes

src/pages/training.js
GBouffard: True.

src/components/Training/course/CourseInfo.js
jsms90: We've removed `constants` from the training page & started using Contentful, straight away :+1: 

As for this `CourseInfo` component, I reckon it requires a few more changes to our content model than is necessarily worth doing within this PR. Raised a Pure Tech Debt ticket here - https://trello.com/c/4dulZp2a/456-review-training-course-content-type

But we've changed the name from `constants`. Let us know what you think

src/pages/training.js
jsms90: @antoniocapelo We figured that `useEffect` wouldn't work for this new `trainingCourseModal` implementation.

Now that each modal is actually it's own page, the transition needs to happen on page load, but [`useEffect` only fires after the page is painted](https://reactjs.org/docs/hooks-reference.html#timing-of-effects)
> Unlike `componentDidMount` and `componentDidUpdate`, the function passed to useEffect fires **after** layout and paint, during a deferred event.

If I'm missing something, and we can still use this, do let me know :+1: 

Assuming that `useEffect` wasn't what we needed, we started trying a different method than React hooks - i.e. [this gatsby plugin](https://www.gatsbyjs.org/packages/gatsby-plugin-page-transitions/). But it seemed prudent to wait until the re-implementation of the modal, which we have already scoped to [a separate ticket](https://trello.com/c/0j78BthW/458-revisit-modal-implementation).

I discussed this with Beatriz and James, who said they were happy for this PR to go in without the animation, as it wasn't part of the original designs anyway. I still think we should put it back in, but just that it might be worth leaving to another ticket? What do you reckon?

tests/createServer.js
```diff
@@ -1,15 +1,12 @@
...
-const { promisify } = require('util')
 const compression = require('compression')
+const { promisify } = require('util')
+const serveStatic = require('serve-static')
+const finalhandler = require('finalhandler')
 
 const compressionHandler = promisify(compression())
 
-const serve = url => {
+const createServer = () => {
```
antoniocapelo: please make it so that this accepts a port and uses it on the listen method

tests/e2e/training-page-test.js
```diff
@@ -1,20 +1,21 @@
...
 require('dotenv').config()
 
-const { GATSBY_ENVIRONMENT } = process.env
-
-const baseUrl =
-  GATSBY_ENVIRONMENT === 'preview'
-    ? process.env.DEPLOY_PRIME_URL
-    : process.env.BASE_URL
-
+const baseUrl = `localhost:3001`
```
antoniocapelo: could you break this like
```js
const hostname = 'localhost'
const port = 3001
const address = `${hostname}:${port}`

```

so that you can pass the port to `createServer`?

src/components/Training/Courses.js
```diff
@@ -24,15 +24,19 @@ const Courses = ({ categories, toggleModal }) => {
...
                 <SectionTitle>{cat.name}</SectionTitle>
               </Padding>
               <Flex column alignStart>
                 {cat.courses.map(course => (
                   <StyledLink
                     style={{ cursor: 'pointer' }}
-                    onClick={() => toggleModal({ ...cat, ...course })}
+                    to={`/training/${cat.slug}/${course.slug}`}
                     key={course.id}
+                    state={{
```
antoniocapelo: according to the [docs](https://github.com/lsirivong/gatsby-plugin-modal-routing#opening-a-page-in-a-modal)

> gatsby-plugin-modal-routing also provides a `Link` component as a convenience to encapsulate this flag for you.

I think that's a cleaner way (instead of passing the state), what do you think?

src/components/Training/course/CourseWrapper.js
```diff
@@ -0,0 +1,47 @@
...
+    transform: scale(1);
+    opacity: 1;
+    visibility: visible;
+    overflow: scroll;
+  `};
+`
+
+const ModalStyles = createGlobalStyle`
+  body {
+    ${is('open')`
```
antoniocapelo: do we need this? 🤔

src/components/Training/course/CourseCloseButton.js
```diff
@@ -0,0 +1,54 @@
...
+    right: ${remcalc(48)};
+  `}
+
+   ${breakpoint('desktop')`
+    right: ${remcalc(90)};
+  `}
+`
+
+const CourseCloseButton = ({ closeTo }) => (
+  <Close to={closeTo}>
```
antoniocapelo: not to be nitpicking but I'm not a big fan of `closeTo` as a prop name. If we're already using a `CloseButton` component, having a "close" preffix on the prop names is a bit unnecessary, don't you think?

src/components/Training/course/CourseWrapper.js
```diff
@@ -0,0 +1,47 @@
...
+      overflow: hidden;
+      position: absolute;
+    `}
+  }
+`
+
+const CourseWrapper = ({ location, content, children }) => (
+  <Layout location={location}>
+    <Wrapper visible={content}>
```
antoniocapelo: do we need the `content` at all? Isn't this only rendered when navigated to?

antoniocapelo: My guess is that if we navigated to this url, `visible` is always true, right? If so, why do we need it?

src/components/Training/Courses.js
```diff
@@ -24,15 +24,19 @@ const Courses = ({ categories, toggleModal }) => {
...
+                      modal: true
+                    }}
+                    data-testid={`modal/${cat.slug}/${course.slug}`}
```
antoniocapelo: I'd change slightly the `data-testid` attribute so it's more meaningful, like: `${course.slug}-link`but that's just a matter of taste

src/components/Training/Courses.js
jsms90: Yeah it's slightly prettier. I guess I just figured that `asModal` vs `state={{modal: true}}` wasn't a significant enough difference, considering the fact that `StyledLink` already uses gatsby's regular Link component under the hood.

So in order to use `asModal` instead, we would be importing an extra component at the top of this file
`import { Link } from 'gatsby-plugin-modal-routing'` & then passing the `as` prop into `StyledLink`. Or, we could import the modal version of Link directly into our `StyledLink` component & have that one decide which one to use under the hood, depending on whether `asModal` is passed?

src/components/Training/course/CourseCloseButton.js
jsms90: sure :+1:

src/components/Training/course/CourseWrapper.js
jsms90: Might be easiest to talk through on a call tomorrow?

tests/e2e/training-page-test.js
```diff
@@ -0,0 +1,44 @@
...
+  })
+  .beforeEach(async t => {
+    firstModalLink = await Selector('a[data-testid^="modal"]').nth(0)
+  })
+  .after(() => server.close())
+
+test('should open the correct training course modal when accessed via a link', async t => {
+  const coursePathName = await firstModalLink
+    .getAttribute('data-testid')
+    .then(data =>
```
antoniocapelo: I think the `data-testid` is useful for targeting specific nodes but they shouldn't contain the logic we're trying to test.

What I mean by this is that using this id for building the course path name doesn't seem right to me.

I'm no QA expert, but the test I'd create for this would be:
- navigate directly to the course link
- save some data about it (URL, page title, etc)
- click 'close
- click on the link for that same course and compare those URL, page title etc to make sure they're the same

src/components/Training/Courses.js
jsms90: I reckon that [the `data-testid` convention](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change/) is quite a nice one, for distinguishing quickly between props which are passed purely for testing purposes vs having any affect on the functionality or styling?

tests/e2e/training-page-test.js
jsms90: I completely agree. I don't like this test relying on how we form our `data-testid`.

The problem that we have is
> navigate directly to the course link

this assumes that we have a course link that's never going to change. Whereas our links are dynamically generated.

What if Greta removes the link that we hard code into this test, from the course catalog in Contentful?

src/components/Training/Courses.js
antoniocapelo: nevermind, I didn't notice it was a `StyledLink`, that causes more work to be made. I'd keep it as it is then :-)

src/components/Training/course/CourseWrapper.js
antoniocapelo: sure!

src/components/Training/Courses.js
antoniocapelo: sure, I'm not saying to remove the prop, I'm just saying that the prop value could be different (probably wasn't explaining myself right=

tests/e2e/training-page-test.js
```diff
@@ -0,0 +1,44 @@
...
+    )
+
+  await t.click(firstModalLink)
+  const location = await getWindowLocation()
+  await t.expect(location.href).contains(`${trainingPageUrl}/${coursePathName}`)
+})
+
+test('should be redirected to when the training course modal is closed', async t => {
+  await t.click(firstModalLink)
+  const modalCloseButton = Selector('a[class^="CourseCloseButton"]').nth(0)
```
antoniocapelo: I'd use a `data-testid` instead of relying on the classname generation logic :-)

antoniocapelo: other than that, the test looks good :)

tests/e2e/training-page-test.js
antoniocapelo: then the tests breaks. Isn't that what we want?

src/components/Training/Courses.js
jsms90: Oh sorry! Gotcha! :+1:

tests/e2e/training-page-test.js
jsms90: Only if the test we are writing is trying to ensure that "node fundamentals" remains on the page as part of our Course Catalog.

If we want to allow Greta to change the catalog to whatever she wants, then we need a test that doesn't care _what_ the course is, but just cares about whether a link on the training page takes the user to a modal relating to that course.

Does that make sense?

antoniocapelo: yes!

src/components/Training/course/CourseInfo.js
```diff
@@ -0,0 +1,47 @@
...
+const CourseInfo = ({
+  name,
+  description,
+  level,
+  preRequisites,
+  preRequisitesCourses,
+  image
+}) => (
+  <Col width={[1, 1, 1, 1, 5 / 6]} px={0}>
+    <Padding bottom={1} data-testid="modal-title">
```
antoniocapelo: isn't this id repeated?

tests/e2e/training-page-test.js
```diff
@@ -0,0 +1,55 @@
...
+})
+
+test("navigating directly to a training course's url should show the same content as navigating via the training page", async t => {
+  await t.click(firstModalLink)
+  const titleFromTrainingPageLink = await Selector(
+    '[data-testid="modal-title"]'
+  ).textContent
+
+  await t.navigateTo('/')
+  await t.navigateTo(`/training/node-js/node-js-fundamentals`)
```
antoniocapelo: instead of navigating directly to this hardcoded URL, I'd capture the location of the page after the click navigation (line 48), and then try to navigate directly to that location. If the CMS ordering changes, this test will fail even though the feature is correct. What do you think?

src/components/Training/course/CourseWrapper.js
```diff
@@ -0,0 +1,35 @@
...
+  opacity: 0;
+  top: 0;
+  left: 0;
+  z-index: ${props => props.theme.zIndexes.modal};
+  transition: transform ${props => props.theme.animations.normal},
+    opacity ${props => props.theme.animations.normal};
+  transform: scale(0.7);
+  visibility: hidden;
+
+  ${is('visible')`
```
antoniocapelo: I wanted to ask about this but forgot: I can't imagine a scenario where the modal has `visible = false`, since it will only render for provided content. In that case, can you try to remove lines 16-24?

tests/e2e/training-page-test.js
jsms90: a phew :+1:

src/components/Training/course/CourseInfo.js
jsms90: oops :flushed:

PR #173, opened 2019-03-26
Fix/dd

antoniocapelo:
## Review template

### DIFF
--- a/src/components/Header/TopNav/Dropdown.js
+++ b/src/components/Header/TopNav/Dropdown.js
@@ -132,9 +132,9 @@ export default class Dropdown extends PureComponent {
         expanded={isExpanded}
         aria-haspopup="true"
         aria-expanded={isExpanded}
-        onFocus={this.handleFocus}
         onClick={this.handleClick}
         onMouseDown={this.handleMouseDown}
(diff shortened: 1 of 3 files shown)


src/components/Header/TopNav/Dropdown.js
```diff
@@ -14,6 +14,7 @@ const DropdownContainer = styled(TopNavItem)`
   position: relative;
   cursor: pointer;
   background: transparent;
+  -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
```
antoniocapelo: Just removing the tap highlight on these topnav items

https://css-tricks.com/snippets/css/remove-gray-highlight-when-tapping-links-in-mobile-safari/

src/components/Header/TopNav/Dropdown.js
```diff
@@ -91,15 +92,35 @@ export default class Dropdown extends PureComponent {
    * so we're detecting it on mouse down
    */
   handleMouseDown = () => {
+    if (this.hasTouch()) {
```
antoniocapelo: This technique should be skipped if we're on a touch deviced — the onclick event is the one that needs to be handled for those cases

PR #174, opened 2019-03-26
Navbar tweaks - change navbar item states' font colour & weight

jsms90:
## Top nav tweaks (2, 3 & 4) - [Trello ticket](https://trello.com/c/askfIg4a/448-top-nav-tweaks)

- switch styles for `hover` & `hoverAndActive` states (on both light and dark theme)
- change the styles that we use for "expanded" state of dropdown select button - now based on "hover" rather than "clickTap", so that it looks the same as the anchors next to it
- add bold font weight to active & active+hover

## Review template

### DIFF
--- a/src/components/Header/TopNav/outerItemStates.js
+++ b/src/components/Header/TopNav/outerItemStates.js
@@ -18,13 +18,11 @@ const clickTapLight = css`
 const activeLight = css`
   background: ${props => props.theme.colors.white};
   color: ${props => props.theme.colors.text};
-  font-weight: bold;
 `
 
 const activeAndHoverLight = css`
   background: ${props => props.theme.colors.greyBG};
   color: ${props => props.theme.colors.text};
-  font-weight: bold;
 `
 
 const defaultDark = css`
@@ -45,13 +43,11 @@ const clickTapDark = css`
 const activeDark = css`
   background: ${props => props.theme.colors.blueBg};
   color: ${props => props.theme.colors.white};
-  font-weight: bold;
 `
 
 const activeAndHoverDark = css`
   background: #3a3553;
   color: ${props => props.theme.colors.white};
-  font-weight: bold;
 `
 
 export const lightStates = {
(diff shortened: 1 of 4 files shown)


antoniocapelo:
flickr
without having looked at the code, I noticed a flickering of the dropdown items, when hovering with the mouse on the dropdown button (see attached gif)

src/components/Header/TopNav/outerItemStates.js
```diff
@@ -17,12 +17,14 @@ const clickTapLight = css`
 
 const activeLight = css`
   background: ${props => props.theme.colors.white};
-  color: ${props => props.theme.colors.textLight};
+  color: ${props => props.theme.colors.text};
+  font-weight: bold;
```
antoniocapelo: I believe this is what's causing the flicker

antoniocapelo: or the other *bold* bellow

src/components/Header/TopNav/outerItemStates.js
```diff
@@ -31,8 +33,8 @@ const defaultDark = css`
 `
 
 const hoverDark = css`
-  background: #3a3553;
-  color: ${props => props.theme.colors.white};
+  background: ${props => props.theme.colors.vibrant};
+  color: #007f56;
```
antoniocapelo: can't we use a theme var here? :-)

src/components/Header/TopNav/outerItemStates.js
```diff
@@ -17,12 +17,14 @@ const clickTapLight = css`
...
+  color: ${props => props.theme.colors.text};
+  font-weight: bold;
 `
 
 const activeAndHoverLight = css`
-  background: ${props => props.theme.colors.text};
-  color: #a9a9a9;
+  background: ${props => props.theme.colors.greyBG};
+  color: ${props => props.theme.colors.text};
+  font-weight: bold;
```
antoniocapelo: IMO, the bold is unnecessary and it really clashes with the rest of the topnav elements :(

src/components/Header/TopNav/outerItemStates.js
jsms90: Is it worth putting it in the theme, if this is the only place in our codebase where we use this colour?

PR #175, opened 2019-03-29
fix bug: Dropdown links not navigating on safari

jsms90:
#### Background
Console logging in Safari shows that `onMouseDown` does not fire on the InnerAnchorItem, when it is clicked.

Our understanding is that, onMouseDown propagates up to DropdownContainer, which changes the `isExpanded` prop (re-added in PR https://github.com/yldio/yld.io/pull/171). When this prop changes, the whole dropdown is re-rendered & navigation doesn't occur. Tbh we're unclear on what is different about Safari from other browsers, to cause this problem.

#### This PR will...
- revert PR https://github.com/yldio/yld.io/pull/172, which removed onMouseDown from InnerAnchorItem
- add `e.stopPropagation` on the inner anchor items, so that `isExpanded: !this.state.isExpanded` doesn't happen on an inner anchor click

## Review template

### DIFF
--- a/src/components/Header/TopNav/Dropdown.js
+++ b/src/components/Header/TopNav/Dropdown.js
@@ -105,6 +105,11 @@ export default class Dropdown extends PureComponent {
     this.setState({ clicked: true, isExpanded: !this.state.isExpanded })
   }
 
+  handleItemClick = e => {
+    e.preventDefault()
+    this.setState({ clicked: false })
+  }
+
   handleFocus = () => {
     if (this.hasTouch()) {
       return
@@ -155,6 +160,7 @@ export default class Dropdown extends PureComponent {
               href={href}
               to={to}
               activeClassName="active"
+              onMouseDown={this.handleItemClick}
             >
               {label}
             </InnerAnchorItem>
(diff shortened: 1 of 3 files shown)


src/components/Header/TopNav/Dropdown.js
```diff
@@ -105,6 +105,12 @@ export default class Dropdown extends PureComponent {
     this.setState({ clicked: true, isExpanded: !this.state.isExpanded })
   }
 
+  handleItemClick = e => {
```
antoniocapelo: I'd call this handleItemMouseDown

antoniocapelo:
I'd suggest this change (see screenshot)

image


This allows for better coordinating the 'show outline' / 'hide outline' logic (eg: currently if we click on the dropdown and then tab through the items, we don't get the outline when reaching the dropdown again — via keyboard). 

Note: I plan to rewrite this component, I'm really not happy with the implementation I did :/

jsms90:
> Note: I plan to rewrite this component, I'm really not happy with the implementation I did :/

Agreed. Could use a refactor :+1: 

Just so we have a record here, we've opened [a new ticket](https://trello.com/c/PQkSm5LK) to address this

PR #177, opened 2019-04-04
feat: color story

antoniocapelo:
## Description
I guess there's no Trello card for this, but this is just me scratching an itch that I had. 

Always felt that having a story for our color variables is easy to do and makes sense. 

Also, I think that having directly the HEX code on the theme variables makes it harder to reason about and to reuse colours for more than one variable, so I moved them outside and named them — as well as I could — based on [this](https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5c61564dd958777e20e8faec).

## Review template

### DIFF
--- a/src/components/Footer/elements.js
+++ b/src/components/Footer/elements.js
@@ -2,7 +2,7 @@ import styled from 'styled-components'
 import { Row } from '../../components/grid'
 
 export const FooterStyled = styled.footer`
-  background: ${props => props.theme.colors.dark};
+  background: #232323;
   color: ${props => props.theme.colors.white};
 `
 
--- a/src/utils/theme.js
+++ b/src/utils/theme.js
@@ -41,6 +41,18 @@ export const zIndexes = {
   gridDebugger: 10001
 }
 
+const colorsVariables = {
+  darkBlue: '#090329',
+  neonGreen: '#65FFCD',
+  darkBlack: '#1d1d1d',
+  black: '#333333',
+  darkGrey: '#828282',
+  lightGrey: '#f9f9f9',
+  grey: '#e6e6e6',
+  white: '#fff',
+  mediumGrey: '#757575'
+}
+
 export default {
   breakpoints,
   spacing,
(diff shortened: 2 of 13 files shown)


src/components/Footer/elements.js
```diff
@@ -2,7 +2,7 @@ import styled from 'styled-components'
 import { Row } from '../../components/grid'
 
 export const FooterStyled = styled.footer`
-  background: ${props => props.theme.colors.dark};
+  background: #232323;
```
antoniocapelo: This is just used here, don't think it makes sense to pollute the theme

src/utils/theme.js
```diff
@@ -41,6 +41,18 @@ export const zIndexes = {
   gridDebugger: 10001
 }
 
+const colorsVariables = {
```
antoniocapelo: here I just renamed the variables according to our zeplin doc (https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5c61564dd958777e20e8faec), however some more colours were added so I named them the best that I could :/

src/utils/theme.js
```diff
@@ -41,6 +41,18 @@ export const zIndexes = {
...
+  darkBlue: '#090329',
+  neonGreen: '#65FFCD',
+  darkBlack: '#1d1d1d',
+  black: '#333333',
+  darkGrey: '#828282',
+  lightGrey: '#f9f9f9',
+  grey: '#e6e6e6',
+  white: '#fff',
+  mediumGrey: '#757575'
```
antoniocapelo: not particularly happy with this name..

src/utils/theme.js
```diff
@@ -57,16 +69,15 @@ export default {
...
-    dark: '#232323',
-    black: '#1d1d1d',
-    lightGrey: '#737272',
-    greyBG: '#f9f9f9',
-    grey: '#e6e6e6',
-    blueBg: '#090329'
+    blueBg: colorsVariables.darkBlue,
+    vibrant: colorsVariables.neonGreen,
+    white: colorsVariables.white,
+    greyBG: colorsVariables.lightGrey,
```
antoniocapelo: secondarytext and light grey were 99% the same color (checked on a color comparison tool, `#757575` vs `#737272`) so I kept just one of them

PR #179, opened 2019-04-08
feat: add space to theme and documented it

antoniocapelo:
## Description 

Having studied a bit the styling libs used in the project (styled-components, styled-system, styled-components-breakpoint, etc), I think they're really very powerful but we're not taking full advantage of them, namely on the `space scale` matter.

I just discovered last week that we have [this](https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5c0019f7e25494019fe389aa) documentation around our spacing rules, which is very handy for our spacing (margin/padding) definitions inside components.

Currently, we're defining spacing with the `styled-components-spacing` lib, which has `Margin` and `Padding` components. While — for some — this may simplify the space definition, I think it creates unnecessary levels of indentation and (worse) adds much more DOM nodes than we'd need. On top of it, we have to create a `spacing` variable in our theme, which got out of hand (a JS object with numbers as keys, and a scale which stopped being a scale at some point).

[https://styled-system.com](https://styled-system.com) handles this spacing — IMO — in a more elegant way, we're you can use style functions to enhance our components with new styling capacities. 

So, for example if you want to style `<div>` element with our `space scale`, you can do:

```js
import styled from 'styled-components'
import { space } from 'styled-system'

const Container = styled.div`
  ${space}
`
// when using it on a render fn, we can do:
<Container ml={1} mr={2} my={3}>
    ... content
</Container>
```

This way our div will have:
- the first scale value as the margin left
- the second as the margin right
- the third as the bottom and top margins

And we can still use responsive styles:
```js
<Container m={[1, 2, 3]}>
    ... content
</Container>
```
This way our div will have:
- the first scale value as the margin for our **first** breakpoint
- the second as the margin for our **second** breakpoint
- the third as the margin for our **third** breakpoint and larger screens

With this, we can document our layout spacings on a story and refer to it every time it's needed.

image

This MR aims to open this discussion so as we can study the possibility of removing the `styled-components-spacing` lib and its usages from the project. 

## Review template

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # YLD Website
 
+
 ## [Link](https://yldio.io/)
 
 ## [Storybook](https://yld-storybook.now.sh)
--- a/src/utils/theme.js
+++ b/src/utils/theme.js
@@ -22,6 +22,17 @@ export const spacing = {
   90: remcalc(90)
 }
 
+export const space = [
+  '0', // 0
+  remcalc(12), // 1
+  remcalc(24), // 2
+  remcalc(36), // 3
+  remcalc(54), // 4
+  remcalc(72), // 5
+  remcalc(108), // 6
+  remcalc(144) // 7
+]
+
 export const breakpoints = {
   smallPhone: 0,
   phone: 471,
@@ -56,6 +67,7 @@ const colorsVariables = {
 export default {
   breakpoints,
   spacing,
+  space,
   elementSizes,
   zIndexes,
   variations: {
--- a/stories/__snapshots__/storyshots.test.js.snap
+++ b/stories/__snapshots__/storyshots.test.js.snap
@@ -7822,6 +7822,298 @@ exports[`Storyshots Image with file url 1`] = `
 />
 `;
 
+exports[`Storyshots Spaces Spaces 1`] = `
+<div>
+  <div
+    style={
+      Object {
+        "alignItems": "center",
+        "display": "flex",
+        "margin": 10,
+        "position": "relative",
+      }
+    }
+  >
+    <p
+      style={
+        Object {
+          "color": "red",
+          "fontSize": 12,
+          "left": "50%",
+          "opacity": 0.5,
+          "position": "absolute",
+          "textAlign": "center",
+          "top": "50%",
+          "transform": "translate(-50%, -50%)",
+          "width": "100%",
+        }
+      }
+    >
+      Level 
+      1
+       - 
+      12
+      px
+    </p>
+    <div
+      style={
+        Object {
+          "background": "#FFE5E5",
+          "height": "0.75rem",
+          "width": 160,
+        }
+      }
+    />
+  </div>
+  <div
+    style={
+      Object {
+        "alignItems": "center",
+        "display": "flex",
+        "margin": 10,
+        "position": "relative",
+      }
+    }
+  >
+    <p
+      style={
+        Object {
+          "color": "red",
+          "fontSize": 12,
+          "left": "50%",
+          "opacity": 0.5,
+          "position": "absolute",
+          "textAlign": "center",
+          "top": "50%",
+          "transform": "translate(-50%, -50%)",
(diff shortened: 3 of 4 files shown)


Daveawb:
I need to add a caveat that you're all aware of already that my React knowledge is in its infancy but from what I do know, using components in this way seems to be massive overkill and if used extensively I can see performance issues coming from it.

In the past I have used SASS to dynamically create CSS that would do this via class attributes. Maybe this is something we could look to use.

(I'm using the variables I have put underneath to generate values)
For example
- Class of `ml-2` will apply `margin-left: 0.5rem`
- Class of 'ml-sm-2` will apply `margin-left 0.5rem` within a media scope for small screen size (see variables).
- Class of `m-auto` will apply `margin-left: auto; margin-right: auto`

The SASS code is below
*Variables*
````
$spacer: 1rem;
$spacers: ();
$spacers: map-merge((
  0: 0,
  1: ($spacer * .25),
  2: ($spacer * .5),
  3: $spacer,
  4: ($spacer * 1.5),
  5: ($spacer * 2),
  6: ($spacer * 3)
), $spacers);

$screen-breakpoints: (
  xs: 0,
  sm: 576px,
  md: 768px,
  lg: 992px,
  xl: 1200px
) !default;
```
*SASS functions for breakpoint management*
```
@function breakpoint-min($name, $breakpoints: $screen-breakpoints) {
  $min: map-get($breakpoints, $name);
  @return if($min != 0, $min, null);
}

@function breakpoint-infix($name, $breakpoints: $screen-breakpoints) {
  @return if(breakpoint-min($name, $breakpoints) == null, "", "-#{$name}");
}

@mixin media-breakpoint-up($name, $breakpoints: $screen-breakpoints) {
  $min: breakpoint-min($name, $breakpoints);
  @if $min {
    @media (min-width: $min) {
      @content;
    }
  } @else {
    @content;
  }
}

*Generating the spacer classes*
```
@each $breakpoint in map-keys($screen-breakpoints) {
  @include media-breakpoint-up($breakpoint) {
    $infix: breakpoint-infix($breakpoint, $screen-breakpoints);

    @each $prop, $abbrev in (margin: m, padding: p) {
      @each $size, $length in $spacers {

        .#{$abbrev}#{$infix}-#{$size} { #{$prop}: $length !important; }
        .#{$abbrev}t#{$infix}-#{$size},
        .#{$abbrev}y#{$infix}-#{$size} {
          #{$prop}-top: $length !important;
        }
        .#{$abbrev}r#{$infix}-#{$size},
        .#{$abbrev}x#{$infix}-#{$size} {
          #{$prop}-right: $length !important;
        }
        .#{$abbrev}b#{$infix}-#{$size},
        .#{$abbrev}y#{$infix}-#{$size} {
          #{$prop}-bottom: $length !important;
        }
        .#{$abbrev}l#{$infix}-#{$size},
        .#{$abbrev}x#{$infix}-#{$size} {
          #{$prop}-left: $length !important;
        }
      }
    }

    // Some special margin utils
    .m#{$infix}-auto { margin: auto !important; }
    .mt#{$infix}-auto,
    .my#{$infix}-auto {
      margin-top: auto !important;
    }
    .mr#{$infix}-auto,
    .mx#{$infix}-auto {
      margin-right: auto !important;
    }
    .mb#{$infix}-auto,
    .my#{$infix}-auto {
      margin-bottom: auto !important;
    }
    .ml#{$infix}-auto,
    .mx#{$infix}-auto {
      margin-left: auto !important;
    }
  }
}

antoniocapelo:
@jsms90 glad we're on the same page! 😄  yeah, this `styled-system` package is really powerful and these `style functions` really let us enhance our components very easily. I'm sad that only recently I got to see the full potential of this, since those `Margin` and `Padding` where in the way :/ 

Anyway, it's no small task getting rid of all those margins and paddings, but it will reduce our DOM nodes count a lot, reduce the file size and indentation so I really hope this gets prioritised :)

PR #180, opened 2019-04-09
fix: add font family styles to storybook

antoniocapelo:
## Description

Since our font family declarations are done in the `html.js` file (to speed up their fetching), they were not being added to the storybook stories. 

This PR fixes this issue.

## Review template

### DIFF
--- a/stories/theme.js
+++ b/stories/theme.js
@@ -1,13 +1,18 @@
-import { ThemeProvider } from 'styled-components'
+import { ThemeProvider, createGlobalStyle } from 'styled-components'
 import React, { Fragment } from 'react'
+
 import theme from '../src/utils/theme'
 import GlobalStyle from '../src/utils/globalStyle'
+const FontsStyle = createGlobalStyle`
+@import url('https://fonts.googleapis.com/css?family=Roboto:400,500,700');
+`
 
 const Theme = storyFn => (
   <ThemeProvider theme={theme}>
     <Fragment>
       {storyFn()}
       <GlobalStyle />
+      <FontsStyle />
     </Fragment>
   </ThemeProvider>
 )


stories/theme.js
```diff
@@ -1,13 +1,18 @@
-import { ThemeProvider } from 'styled-components'
+import { ThemeProvider, createGlobalStyle } from 'styled-components'
 import React, { Fragment } from 'react'
+
 import theme from '../src/utils/theme'
 import GlobalStyle from '../src/utils/globalStyle'
+const FontsStyle = createGlobalStyle`
+@import url('https://fonts.googleapis.com/css?family=Roboto:400,500,700');
```
jsms90: How about we put FontStyle inside `src/utils` at the same level as `globalStyle`, or even within the same file?

Then we can use the same fonts component in Storybook and in our app, inside the render function in Layout.js:
https://github.com/yldio/yld.io/blob/9eccb11214b789b0d2742429741081449b49a519/src/components/layout.js#L104-L106

Right now, I think we get our fonts into our website via
- [a stylesheet in `html.js`](https://github.com/yldio/yld.io/blob/9eccb11214b789b0d2742429741081449b49a519/src/html.js#L27)
- [this `font-family` line in globalStyle](https://github.com/yldio/yld.io/blob/9eccb11214b789b0d2742429741081449b49a519/src/utils/globalStyle.js#L11)

From our chat with @Daveawb about what may (or may not) be coming up in future, with regards to events pages/components on yld.io, we may need to alter the font on a page by page basis in future. So this might be worth doing now?

antoniocapelo: @jsms90 I don't think that change (moving it to src/utils and using it on `Layout.js`) should be done. The reason we have these two lines in our `Layout.js`

```
        <link
          crossOrigin="true"
          rel="preconnect"
          href="https://fonts.gstatic.com/"
        />
        <link
          href="https://fonts.googleapis.com/css?family=PT+Mono|Roboto:400,500,700"
          rel="stylesheet"
        />
```
is because of performance reasons. Having the fontfamily being `@imported` on the `css` files causes a delay on the fetch requested for that stylesheet. Then, there's a new connection to `https://fonts.gstatic.com` for downloading the font files themselves. 

Putting these two `link` tags starts the download of the stylesheet as soon as possible and `preconnects` (creates the initial connection) to https://fonts.gstatic.com which also saves some time.

However, Storybook is not aware of this `Layout.js` file nor has a way to include these links in its `head` tag. This is way I'm `@importing` the font stylesheet only on the storybook (since performance is not an issue there).

Do you have a better solution for this?

jsms90: Fair enough.

No, no suggestions on doing this differently for storybook.

PR #181, opened 2019-04-10
Feat/visual regression  [WIP]

antoniocapelo:
## Description - [Trello ticket](https://trello.com/c/gLhIK9bK)

Summary of the current state of this branch:
- found correct version of `addon-storyshots-puppeteer` that correctly runs the tests agains our storybook version(we should try to update our storybook to v5 though..)
- Tests are running and we're getting failures when expected 🙌
- Tests are running fine against the 'live' storybook version (`yarn storybook`) but I'm having troubles trying to run the tests against the static build of storybook (`yarn build:storybook`), mainly on the stories that use images (they seem to fail to load)
- When the components have an initial animations (like the blur-up of the image on the Case Study Preview or the logo.gif animation) we're having false negatives because the screenshot is being taken at different times on each run. I'm currently using the `beforeScreenshot` function option of the addon so that it waits before taking the screenshot and let the animations end.. One possible (not perfect) improvement would be to decide to wait or not based on the story information we receive on the `beforeScreenshot` callback. 

## Review template


omonk:
Closing this in favour of clearing from open PRs and returning to it in near future

PR #184, opened 2019-04-10
Fix:external links to open new tabs

GBouffard:
## Description - [375](https://trello.com/c/AuoqvQlI/375-make-all-external-links-to-open-on-a-new-tab)

User stories/acceptance test:
- As a user when I navigate to the `About Us Page` and click the `AddCode link`, I expect the link to open the external site **in a new tab**.
- " " to the `SEO pages` " " and click `more talks` link " " " a new tab.
- " " to the `Node SEO page` " " and click `more books` link " " " a new tab.
- " " to the `Join Us page` " " and click `Git Hub` link " " " a new tab.
- " " to the `Join Us page` " " and click `StackOverflow` link " " " a new tab.

Engineering notes:
"Git Hub" and "StackOverflow" (on Join Us Page) - are coming from a contentful text in which they are encoded. The component rendering didn't have a className or tagName making it easy to grab (especially if the content of the page changes). What was done was to grab all anchors from the 'Join us' page and filter by the one with a title attribute (the only ones with a title at the moment are our 2 links coming in contentful). the target _blank attribute is then applied to these 2; if new links are added or updated from Contentful it should still work.

Refactoring and improvements:
- "more books" (on Node SEO page) - both the text and the link are hardcoded within the code. the text "More books" will work on any topic but the link should at least come from Contentful.
- Have the title bodyPrimary text of the BooksSection utilising the SectionTitle title coming from Contentful
- refactor TitleAndList.js component to bypass eslint react/no-display-name rule (assign to a Comp constant and return it. Better to return directly and use  eslint-disable-next-line instead. I checked with Capelo that it's a valid refactor)

## Review template


antoniocapelo:
@GBouffard if you look into the code of the `StyledLink` component you'll find an `external` prop, which indicate that is an external link (that by default opens in a new tab). So no need to add the `target` attr, just use that prop and the component does that for you

src/components/AboutUs/Subsidiaries.js
```diff
@@ -16,7 +16,7 @@ const Subsidiary = ({ image, description, linkUrl, linkText }) => (
         {description}
       </BodyPrimary>
       {linkText ? (
-        <StyledLink reverse href={linkUrl}>
+        <StyledLink reverse href={linkUrl} target="_blank">
```
jsms90: Our `StyledLink` component takes a prop called `external`, which adds `target="_blank"` but also `rel='noopener noreferrer'`
https://github.com/yldio/yld.io/blob/3ec6b582c7f8c3735559674d15284b867e5b745f/src/components/Common/StyledLink.js#L87-L107

([This site](https://support.performancefoundry.com/article/186-noopener-noreferrer-on-my-links) offers a good explanation of what noopener & noreferrer are all about)

So could we replace these `target="_blank"` with `external` :pray: ?

---

@Daveawb @antoniocapelo As a slight aside, we could make it easier for new developers to use our existing library of components, if we had more thorough documentation (i.e. storybook)? For this situation in particular, we could document the `external` prop more effectively, using the [links addon ](https://github.com/storybooks/storybook/tree/next/addons/links) & the [info addon](https://github.com/storybooks/storybook/tree/next/addons/info)?

Given the rotating nature of the staffing on this project, I would suggest that we start looking into a more comprehensive way of documenting our components than just the basic storybook functionality (e.g. additional info, like this). A combination of "info" and ["knobs"](https://github.com/storybooks/storybook/tree/next/addons/knobs) would be a good start in enabling us to explain/demonstrate more to devs.

If anyone has the time, [here's an article on using React + storybook in the context of a design system](https://medium.com/buildit/using-react-within-a-design-system-73d4bb0cc822)

src/components/Speciality/Books.js
```diff
@@ -10,6 +10,9 @@ import breakpoint from 'styled-components-breakpoint'
 import { SectionTitle, CardTitle, BodyPrimary } from '../Typography'
 import StyledLink from '../Common/StyledLink'
 
+// TODO - should come from contentful
+const moreBooksLink = 'http://nodepatternsbooks.com/index.html'
```
jsms90: I reckon any time we add a TODO to the codebase, we should document it in a trello ticket. What do you guys think?

In this situation, I guess it could go into a checklist in the content model review?

src/components/Speciality/Books.js
```diff
@@ -54,7 +57,7 @@ const BooksSection = ({ books, title }) =>
           <CenteredCol width={[1]}>
             <SectionTitle small>{`${title.trim()} books`}</SectionTitle>
             <BodyPrimary>
-              Node.js books written by members of YLD for our community.
+              {`${title.trim()} books written by members of YLD for our community.`}
```
jsms90: Nice :+1:

I wonder whether we want this line to be editable by Greta? :thinking:

src/components/Speciality/Books.js
GBouffard: Agreed 👍

src/components/AboutUs/Subsidiaries.js
GBouffard: Agreed. New comers to the site need to know what is available.

src/components/Common/TitleAndList.js
```diff
@@ -22,6 +22,13 @@ const TitleAndList = ({
   themeVariation = 'white',
   extraContent
 }) => {
+  useEffect(() => {
+    const titledAnchors = Array.from(document.querySelectorAll('a')).filter(
+      a => a.title
+    )
+    titledAnchors.forEach(a => a.setAttribute('target', '_blank'))
+  })
+
```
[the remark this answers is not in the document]
GBouffard: Pushed the change 👍

src/components/Speciality/Books.js
GBouffard: The line being hard-coded tells us that we only use Books.js for our Node.js speciality so I would not worry about making it editable at this point in time; but could be it could be the case in the future.

src/components/AboutUs/Subsidiaries.js
Daveawb: Yes, this documentation is absolutely necessary.

src/components/Common/TitleAndList.js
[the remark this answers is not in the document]
Daveawb: It'd be a good idea to unit test this

src/components/AboutUs/Subsidiaries.js
GBouffard: Do we have a Trello ticket where we could add these notes? We should create one.

src/components/Common/TitleAndList.js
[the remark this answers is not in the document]
GBouffard: To unblock the merge, shall I add this to the unit test Trello ticket (464/468) which I started to look into?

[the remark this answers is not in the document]
Daveawb: Yeah, don't do it here.

PR #185, opened 2019-04-11
fix: Sort events array by date

GBouffard:
## Description - [479](https://trello.com/c/m3C3DpRH/479-meetup-events-component-events-should-be-in-order-of-date)

This fix sort the meet-up events array by date

Although we only display from the actual date on the homepage (and there is only 1 future event at the moment), past events are still part of the events array passed to the component so it's testable. I added a 1-line function that guaranty the sort order by date within `getEvents.js getHomepageMeetups` function.
Adding console.log(events.sort((a, b) => (a.node.date < b.node.date ? -1 : 1))) at the top of getHomepageMeetups function returns the array of events sorted by date, which wasn't the case before the fix.

## Review template


src/components/Homepage/events/getEvents.js
```diff
@@ -2,6 +2,7 @@ import { format, isAfter } from 'date-fns'
 
 export const getHomepageMeetups = events =>
   events
+    .sort((a, b) => (a.node.date < b.node.date ? -1 : 1))
```
antoniocapelo: what's the var type of the `node.date` property? is it a string or number? if a number, I agree with this change, but if it's a string you'd be making a string comparison which would give you wrong results.

I'm seeing the line 7 we're doing `new Date(n.node.date)` so I'm just making sure we're comparing the correct types.

Finally, and this is a small improvement, we should sort after the filtering, so instead of sorting and reducing the collection size, we can only sort on the necessary items :)

GBouffard: That's true. Let me double-check and update if needed.

src/components/Homepage/events/getEvents.js
```diff
@@ -1,9 +1,12 @@
 import { format, isAfter } from 'date-fns'
 
+const toDate = stringDate => new Date(stringDate)
+
 export const getHomepageMeetups = events =>
   events
     .filter(n => !n.node.homepageFeatured)
     .filter(n => isAfter(new Date(n.node.date), new Date()))
+    .sort((a, b) => (toDate(a.node.date) < toDate(b.node.date) ? -1 : 1))
```
Daveawb: Minor edge case: if `a.node.date` and `b.node.date` are equal, this will always return `-1` meaning that if an update occurred that causes a re-render of the component these are stored in, the values will switch places over and over. try to always use `<=` or `>=` in comparator functions / lambdas.

GBouffard: Great point. I didn't think of it but it would switch them for sure. I used `<=`. Thanks for  pointing it out

PR #186, opened 2019-04-12
use same JobLink component on homepage & Join Us page (entire component is clickable)

jsms90:
## Description - [Trello ticket](https://trello.com/c/YqhNl7xe/300-joblink-component)

- took Join Us page version, where whole component is a clickable link, moved this to Common, removed usage of `Padding` & altered prop names
- used new `Common/JobLink` component on homepage
- used new `Common/JobLink` component on Join Us page
- created JobLink story
- updated snapshots

I would have liked to refactor more as I went. But since I'm going on holiday soon & we have meetings, I just settled for the minimum necessary to complete the ticket.

## Review template


src/components/Common/JobLink.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import Flex from 'styled-flex-component'
+import styled from 'styled-components'
+
+import ExternalAnchor from '../Common/ExternalAnchor'
+import { BodyPrimary } from '../Typography'
+import Hr from '../Common/Hr'
+
+const StyledExternalAnchor = styled(ExternalAnchor)`
+  height: 107px;
```
antoniocapelo: is this needed? shouldn't it be a min-height? (Im afraid of content being cut).
Also, we should use `remcalc`?

jsms90: Yeah good point. I was fixing the height cos that's in the specs:
image

But this is a lazy half-done implementation cos I'm rushing. I'll put it back to how it was cos I don't think I can add ellipses, or some other solution to cut of the text, before I leave.

PR #190, opened 2019-04-29
Implement LogoGrid Zeplin specs on live website

jsms90:
## Description - [Trello ticket](https://trello.com/c/JSh3ln1T/379-logogrid-fixes)

Zeplin documentation for this component was created after the 'logo grids' on our website were built. This PR brings the existing pages in line with the Zeplin spec. It involves alterations to:
1. Statement/Companies section of the homepage
2. Projects section of all Speciality (aka SEO) pages
3. Technology Partnerships section of About Us page

This PR will:
- use the old `Companies` component from the homepage as the basis for a new `LogoGrid` component, now inside the Common folder
- alter this component to have 3 columns on small tablet (used to have 4 columns)
- add 12px spacing between each row of logos
- re-use this component on the Speciality template
- make the 'Other companies we helped' section of the Speciality template optional (i.e. remove this subtitle if/when there are no logos entered into Contentful) - fixes [another bug ticket](https://trello.com/c/0Xl8J07p/492-other-clients-section-of-speciality-page-sometimes-has-no-logos-but-still-has-subtitle-showing)
- re-use this component on the About Us page for the Technology Partnerships section, so that the same column spacing rules apply there  
(Should also fix [this bug ticket](https://trello.com/c/KMY5ABt5/488-logogrid-bug-on-about-us-page))

## Review template

### DIFF
--- a/src/components/Common/LogoGrid.js
+++ b/src/components/Common/LogoGrid.js
@@ -11,31 +11,27 @@ const Column = styled(Col)`
   padding-bottom: ${props => props.theme.spacing[1]};
 `
 
-const LogoGrid = ({ logos = [] }) =>
-  logos ? (
-    <Row>
-      {logos.map(logo => (
-        <Column
-          width={[1 / 2, 1 / 2, 1 / 2, 1 / 2, 1 / 3, 1 / 4]}
-          key={logo.id}
-        >
-          {logo.url ? (
-            <ExternalAnchor href={logo.url}>
-              <Image
-                image={logo.image}
-                width="250px"
-                style={{ filter: 'grayscale(1)', saturate: '0' }}
-              />
-            </ExternalAnchor>
-          ) : (
+const LogoGrid = ({ logos }) => (
+  <Row>
+    {logos.map(logo => (
+      <Column width={[1 / 2, 1 / 2, 1 / 2, 1 / 2, 1 / 3, 1 / 4]} key={logo.id}>
+        {logo.url ? (
+          <ExternalAnchor href={logo.url}>
             <Image
-              image={logo}
+              image={logo.image}
+              width="250px"
               style={{ filter: 'grayscale(1)', saturate: '0' }}
             />
-          )}
-        </Column>
-      ))}
-    </Row>
-  ) : null
+          </ExternalAnchor>
+        ) : (
+          <Image
+            image={logo}
+            style={{ filter: 'grayscale(1)', saturate: '0' }}
+          />
+        )}
+      </Column>
+    ))}
+  </Row>
+)
 
 export default LogoGrid
--- a/src/components/Common/LogoGrid.js
+++ b/src/components/Common/LogoGrid.js
@@ -10,18 +10,18 @@ const Column = styled(Col)`
   padding-bottom: ${props => props.theme.spacing[1]};
 `
 
-const Companies = ({ companies = [] }) =>
-  companies ? (
+const LogoGrid = ({ logos = [] }) =>
+  logos ? (
     <Row>
-      {companies.map(company => (
+      {logos.map(logo => (
         <Column
           width={[1 / 2, 1 / 2, 1 / 2, 1 / 2, 1 / 3, 1 / 4]}
-          key={company.id}
+          key={logo.id}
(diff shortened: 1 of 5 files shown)


src/components/Common/LogoGrid.js
```diff
@@ -0,0 +1,41 @@
...
+import ExternalAnchor from '../Common/ExternalAnchor'
+
+const Column = styled(Col)`
+  max-height: 108px;
+  display: flex;
+  align-items: center;
+  padding-bottom: ${props => props.theme.spacing[1]};
+`
+
+const LogoGrid = ({ logos = [] }) =>
```
GBouffard: we shouldn't have a default value with the ternary running below.
the boolean value of [] returns true so null would never run. Remove the `= []`

src/components/Speciality/Projects.js
```diff
@@ -83,15 +83,10 @@ const ProjectsSection = ({ related, title, clients }) => {
...
-        <CompaniesHelped noOther clients={clients} />
-      </Padding>
-    </Grid>
-  )
-}
+      ) : null}
+      {clients ? <CompaniesHelped clients={clients} /> : null}
+    </Padding>
+  </Grid>
+)
```
GBouffard: Very nice refactoring with it 👍

src/components/Common/LogoGrid.js
```diff
@@ -0,0 +1,41 @@
...
+            <ExternalAnchor href={logo.url}>
+              <Image
+                image={logo.image}
+                width="250px"
+                style={{ filter: 'grayscale(1)', saturate: '0' }}
+              />
+            </ExternalAnchor>
+          ) : (
+            <Image
+              image={logo}
```
GBouffard: Is it not supposed to also be `image={logo.image}` here?

jsms90: Unfortunately not...but I agree that this looks weird :thinking: The discrepency is because the way logos are added to a section is different depending on the page

It's because, if there is a url associated with the logo (e.g. on the [About Us page](https://github.com/yldio/yld.io/blob/master/src/pages/about-us.js#L161-L173)), then the shape of the object from Contentful comes from this graphql query:
```
logos {
            name
            url
            image {
              title
              file {
                url
              }
              fluid(maxWidth: 250) {
                ...GatsbyContentfulFluid_withWebp
              }
            }
          }
```

Otherwise, as in the [speciality template](https://github.com/yldio/yld.io/blob/master/src/templates/speciality.js#L164-L171), the `logo` = `image` only. I.e. this is the shape:
```
logos {
        id
        title
        file {
          url
          fileName
        }
      }
```

Unsure what the best way is to make this cleaner... :thinking:

src/components/Common/LogoGrid.js
jsms90: Good point. I've just copied this lazily from the original.

I'm not a big fan of the pattern where a component either renders itself or renders null. My personal preference would be for the parent to decide whether to show/hide the child or not.

For example, if `logos` don't exist in Contentful for one of the speciality pages, we don't just want the LogoGrid to disappear. We want the whole section (including the associated subtitle) to disappear. This wasn't happening previously, because we were allowing the LogoGrid to show/hide itself rather than handling it at the parent level.

I'll remove this ternary altogether :+1:

GBouffard: i agree: In theory the parent should decide. Nice.

src/components/Common/LogoGrid.js
GBouffard: Wow. As misleading as it is refactoring is not a priority - we have other things to focus on but thanks for the explanation. It is for sure confusing for people who don't know.

One thing we should do because it should not mean too much extra work is document these shapes in `Readme`s for future developers that join the project.

jsms90: We chatted about this via Slack, but https://github.com/yldio/yld.io/pull/190/commits/a6ccb1ece84e3ba27a52954aaa51f67f54b32a9d seems to provide sufficient clarity for now.

Going to add the readme addon to storybook under another ticket though. Good to have either way.

PR #192, opened 2019-04-30
update blog and tutorial sections to be consistent

jsms90:
### What
I have (mostly) followed Zeplin specs for ["Blog listing" component](https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5cadbea6b962339fc2452c9e) to make the following consistent:
- Blog section on homepage
- Blog section on speciality template
- Tutorials section on speciality template

In the process, I have also changed the CTA text to "Visit our blog" (ticket 3)

Note: [On some speciality pages (e.g. React)](www.yld.io/speciality/react-js) this component sits next to the existing `GetInTouch` component, which has different specs for the columns on various breakpoints.
So, for the sake of consistency across our website, I have deferred to the GetInTouch specs over the BlogListing spec in Zeplin for now.

### How
All 3 of the above sections are built from a new component called `TitleAndMediaList`, imagining that 'media' could mean either a blog or a tutorial. This generic component is useful for keeping the layout identical, no matter the content.

Another Common component called `BlogListing` is built on top of the generic `TitleAndMediaList` & is also responsible for converting graphql content from Medium, into the right format for presentation.

Both `TitleAndMediaList` & `BlogListing` are re-used & therefore in the Common folder + have their own stories. The tutorial version, however, exists only once & therefore remains inside the Speciality folder.

## Review template


src/components/Common/BlogListing.js
```diff
@@ -0,0 +1,35 @@
...
+import { Padding } from 'styled-components-spacing'
+import { format } from 'date-fns'
+
+import { Grid } from '../grid'
+import TitleAndMediaList from '../Common/TitleAndMediaList'
+
+const BlogListing = ({ title, description, posts }) => {
+  const mediaItems = posts.map(
+    ({ id, title, uniqueSlug, firstPublishedAt }) => {
+      return {
```
GBouffard: a minor thing but you can remove return keyword `=> ({ id, title, href`

src/components/Speciality/Tutorials.js
```diff
@@ -1,55 +1,37 @@
...
+import { Grid } from '../grid'
+import TitleAndMediaList from '../Common/TitleAndMediaList'
 
-const TutorialsGrid = styled(Grid)`
-  padding-bottom: ${remcalc(18)};
-`
+const TutorialsSection = ({ tutorials }) => {
+  const mediaItems = tutorials
+    .slice(0, 3)
+    .map(({ id, title, link, additionalInfo }) => {
```
GBouffard: same comment as above to remove unneeded return keyword.

src/components/Common/BlogListing.js
```diff
@@ -0,0 +1,35 @@
...
+  )
+
+  return (
+    <Grid>
+      <Padding vertical={{ desktop: 5, smallPhone: 3.5 }}>
+        <TitleAndMediaList
+          title={title}
+          description={description}
+          mediaItems={mediaItems}
+          CTALink="https://medium.com/yld-engineering-blog"
```
GBouffard: I would store `https://medium.com/yld-engineering-blog` in a const since it's being used twice in the file.
A side note, not a requested change but CTALink / CTAText should come from Contentful.

src/components/Speciality/Tutorials.js
```diff
@@ -1,55 +1,37 @@
...
-            </Col>
-          </Row>
+      <Grid>
+        <Padding vertical={{ desktop: 5, smallPhone: 3.5 }}>
+          <TitleAndMediaList
+            title="Tutorials"
+            description="NodeJS tutorials created by members of YLD for the community."
+            mediaItems={mediaItems}
+            CTALink="http://nodetuts.com/"
+            CTAText="More tutorials"
```
GBouffard: not a requested change but hard-coded strings/descriptions should come from contentful.

jsms90: Yeah I completely agree that these should come from Contentful. But I would suggest that we consider alterations to Contentful to be out of the scope of this ticket.

If it were just a matter of adding a couple of new fields, then we could definitely do this at the same time :+1:

But we don't currently have a content type on Contentful, which is for entering information which is to be used across all Speciality pages. Things that would fall under this category include the CTALink & CTAText for this blog section, but also the section titles & section descriptions for many of the other sections on these pages.

Right now, most of these things are hard-coded into the template. The only things in Contentful are the page-specific things (i.e. the ones that change depending on whether it's a React page or a Node page).

So, in order to do what you're suggesting, we would ideally be creating a new content type for the Speciality _Template_ itself - things to be shared across all speciality pages. Rather than creating a whole new content type right now, just for these 2 strings, I would suggest that we wait until we can do it in a more systematic and thorough way, that accounts for all of the common elements - basically when Ollie joins us.

src/components/Common/BlogListing.js
jsms90: I see what you mean. But just because they happen to be the same, doesn't necessarily mean that they are conceptually always supposed to be identical.

For example, Greta could theoretically change the CTALink so that it directed people to the "latest stories", i.e. https://medium.com/yld-engineering-blog/latest
But we would never want the text appended to the beginning of the slug to change from https://medium.com/yld-engineering-blog

Anyway, when we do have the CTALink coming from Contentful, I think line 27 will become `CTALink={CTALink}` but line13 (the `href`) will remain hard-coded in our codebase :woman_shrugging:

GBouffard: No worries. if they may not always be identical then no need to store in a const 👍

src/components/Speciality/Tutorials.js
GBouffard: That makes sense. Let's talk about it when Ollie is with us 👍

PR #195, opened 2019-05-08
Yld-logo-svg-colour-changes

GBouffard:
## Description - [500](https://trello.com/c/y1Jp1QNd/500-yld-logo-svg-colour-changes)

- [x] Change the colour of the logo square on the Engineering/Design/Training pages to be #333333
- [x] Change the colour of title (e.g Engineering) on the Engineering/Design/Training pages to be #333333
- [x] Amend the SEO page logo squares to their respective illustrations colours:
respectively:
Node.js: #52FFAC
Graphql: #EB008B
Vue: #039328
React: #0BDDF9
Kubernetes: #29EFEF
- [x] Refactored a bit the code to be more explicit

## Review template


src/components/Header/TopNav/ServicesSpecialitiesMap.js
```diff
@@ -7,34 +9,43 @@ const specialitiesMap = {
...
+    graphql: '#EB008B',
+    'vue-js': '#039328',
+    'react-js': '#0BDDF9',
+    kubernetes: '#29EFEF'
+  },
+  design: {},
+  training: {},
+  delivery: {},
+  dedicatedTeams: {},
+  openSource: {},
```
omonk: do we need any of these?

GBouffard: Not yet - but this was part of a previous ticket to have these services defined somewhere.
So far only the first 3 are live on our website (and only engineering has specialities pages. They all will have in the future)

src/components/Header/TopNav/ServicesSpecialitiesMap.js
```diff
@@ -7,34 +9,43 @@ const specialitiesMap = {
...
-  engineering: '#52FFAC',
-  design: '#52FFAC',
-  training: '#52FFAC',
-  delivery: '#52FFAC',
-  dedicatedTeams: '#52FFAC',
-  openSource: '#52FFAC',
-  specialityText: '#090329',
-  specialityHover: 'white'
+  engineering: {
+    'node-js': '#52FFAC',
```
jsms90: The `logoColor` for each speciality page is a value which comes from Contentful. So I reckon it's probably not ideal to be hard-coding them in here & duplicating the "source of truth", as it were.

We also want to be getting to a point where Greta can add new speciality pages without consulting an engineer. So ideally, we wouldn't have the list of specialities/services hard-coded into the codebase either.

That one is a little more work, given that the TopNav now relies on `specialitiesMap`, so can probably be dealt with as part of a separate ticket. But off the top of my head, one solution would be to move the navbar outside of `Layout`, since this component needs to change its contents, depending on which page its on.

GBouffard: I didn't know; but now that I do, it makes perfect sense. You are right this should probably be part of another ticket. Shall I create one or do you want to? Once this is done is this PR ready to go?

jsms90: Well there's 2 things here:
- the logo colours are now being hard-coded (part of this PR)
- this file hard-codes the names of the specialities (can be part of another ticket)

jsms90: Oh no. My bad, both are already hard-coded on master & this PR is just moving the location into an object called `engineering`.

Yeah, it can all go in a separate ticket then :+1:

PR #197, opened 2019-05-09
Fix tutorials section bug

jsms90:
## Description - [Trello ticket](https://trello.com/c/IvgD4Pmh/502-tutorials-section-is-appearing-on-speciality-pages-where-there-are-no-tutorials)

- Remove TutorialsSection from speciality pages that don't have any associated tutorials
- Fix description text, so that 'NodeJS' is no longer hard-coded into this section, regardless of which speciality page it's on

## Review template

### DIFF
--- a/src/templates/speciality.js
+++ b/src/templates/speciality.js
@@ -27,6 +27,7 @@ const Speciality = ({
   },
   location
 }) => {
+  console.log('speciality', speciality)
   const {
     relatedProjects,
     title,
@@ -40,7 +41,9 @@ const Speciality = ({
   } = speciality
 
   const posts = filteredPosts.map(({ node }) => node)
+  const talks = getExternalType(speciality, `Talk`)
   const tutorials = getExternalType(speciality, `Tutorial`)
+  const books = getExternalType(speciality, `Book`)
 
   return (
     <Layout backgroundColor="blue" location={location}>
@@ -59,10 +62,7 @@ const Speciality = ({
         title={title}
       />
       <EventSection events={events} title={title} eventIcon={eventIcon} />
-      <TalksSection
-        talks={getExternalType(speciality, `Talk`)}
-        videoIcon={videoIcon}
-      />
+      <TalksSection talks={talks} videoIcon={videoIcon} />
       <BlogListing
         title="Blog posts"
         description={`${title} articles created by members of YLD for the community.`}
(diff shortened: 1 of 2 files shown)


src/templates/speciality.js
```diff
@@ -58,17 +62,16 @@ const Speciality = ({
...
-      />
+      <TalksSection talks={talks} videoIcon={videoIcon} />
       <BlogListing
         title="Blog posts"
         description={`${title} articles created by members of YLD for the community.`}
         posts={posts}
       />
-      <TutorialsSection tutorials={getExternalType(speciality, `Tutorial`)} />
-      <BooksSection title={title} books={getExternalType(speciality, `Book`)} />
+      {tutorials.length > 0 ? (
```
omonk: need to check first that tutorials is no undefined otherwise we could be seeing `cannot read property length of undefined`

`{tutorials && tutorials.length && ....}`

src/templates/speciality.js
```diff
@@ -27,6 +27,7 @@ const Speciality = ({
   },
   location
 }) => {
+  console.log('speciality', speciality)
```
omonk: remove

omonk: has there been any previous discussion of adding no console log rules to eslint?

jsms90: There hasn't been any previous discussion. I thought it was already in there as a rule :confused:

src/templates/speciality.js
jsms90: Because of [this line](https://github.com/yldio/yld.io/pull/197/files/2a9681fb0e7255d247ddafe577eb4b2e0ea2caef#diff-6a7ed84b5bdb2c4809423a71f05a42eeL20) there will actually always be an empty array. But you're right, probably a good idea to code more defensively in general.

Thanks :+1:

GBouffard: Ok, so there is already a safety net with the line your mentioned.
One minor change I'd suggest. You don't need to compare to 0.
`tutorials.length ? (` is enough for your ternary

src/templates/speciality.js
GBouffard: I thought so as well. Has it been removed?

src/templates/speciality.js
jsms90: Yeah it's true. I just think `tutorials.length > 0` reads nicely :laughing:

PR #198, opened 2019-05-10
Speciality-pages-events-section

GBouffard:
## Description - [493](https://trello.com/c/voeMZxoC/493-blue-sections-on-speciality-pages-are-touching-when-light-section-between-them-is-optional-missing)

- [x] filter events from today's date + by speciality
_nb: the actual filter that was in place was only filtering from date. There do not seem to be tags against an event so the filter I put in place looks for the speciality name within each event.eventTitle (case insensitive). If there is a better way, I'm open to suggestions._
- [x] display list upcoming events within the relevant speciality
- [x] display a simple message and CTA in the event that there are no events
- [x] Unrelated to the task but related to specialities and the task I did before: Resolve React warning in ServiceLink component (custom prop color of type string instead of passing a boolean)
- [ ] style following Zeppelin

## Review template

### DIFF
--- a/src/components/Speciality/Events.js
+++ b/src/components/Speciality/Events.js
@@ -55,7 +55,7 @@ const EventSection = ({ events, title, eventIcon }) => {
               <Padding bottom={1}>
                 <img src={specialityEventIcon} alt="events icon" />
               </Padding>
-              <SectionTitle>{`Upcoming ${title} events`}</SectionTitle>
+              <SectionTitle>{`Upcoming ${title.trim()} events`}</SectionTitle>
             </div>
           </Col>
           <Col width={[1, 1, 1, 1, 6 / 12]}>
--- a/src/components/Speciality/Events.js
+++ b/src/components/Speciality/Events.js
@@ -6,7 +6,19 @@ import specialityEventIcon from './assets/events-icon.svg'
 import { Row, Col, Grid } from '../grid'
 import { SectionTitle, Subtitle, BodyPrimary } from '../Typography'
 import ExternalAnchor from '../Common/ExternalAnchor'
+import StyledLink from '../Common/StyledLink'
 import Hr from '../Common/Hr'
+
+const CTA = {
+  link: {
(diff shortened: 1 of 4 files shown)


src/components/Speciality/Events.js
```diff
@@ -1,56 +1,71 @@
...
 import ExternalAnchor from '../Common/ExternalAnchor'
-import { Padding } from 'styled-components-spacing'
-import { format } from 'date-fns'
+import Hr from '../Common/Hr'
 
-const EventBorder = styled(Col)`
-  border: 1px solid rgba(51, 51, 51, 0.15);
-`
+const noEventsMessage = title =>
+  `It looks like there currently aren’t any upcoming ${title} events. You can always check back again later or get in touch if you are interested in potentially hosting one.`
```
omonk: is this copy signed off?

src/components/Speciality/Events.js
```diff
@@ -1,56 +1,71 @@
...
-                      </Subtitle>
-                      <BodyPrimary noPadding>
-                        {format(new Date(event.date), 'MMMM DD[,] dddd')}
-                      </BodyPrimary>
-                    </Col>
-                  </Row>
-                </Padding>
-              </EventBorder>
-            ))}
+            {specialityEvents.length ? (
```
omonk: `{specialityEvents && specialityEvents.length && ....` specialityEvents could be null without `length` property

src/components/Speciality/Events.js
GBouffard: Yes. See on Zeppelin:
https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5cd2acb3eadfea34c03e671e

src/components/Speciality/Events.js
GBouffard: In theory it shouldn't as I added a fall back so it equals to `[]`. I have a few updates to push so the code is more robust.

PR #199, opened 2019-05-10
C fix meetup lambda

AniaMakes:
## Description - [Trello ticket](https://trello.com/c/KN345xF2/478-no-events-should-be-missing-on-the-meetup-events-component)

In this repo
- updated env variable names to be consistent
- fixed the Contentful calls (I believe the API has changed slightly since the function was first written)
- added homepageFeatured to generateContentfulEvent as this is a compulsory field
- added a status code to the lambda
- added webpack-node-externals so that the backend modules associated with this lambda can be bundled neatly
- added comments on order of Contentful calls order for easier debugging in the future
- added note on Personal Access Token to the readme

Outside of repo but related
- created (via JP) a joint google account to add to contentful so that there is a joint entity for Personal Access Tokens
- added the PAT to netlify secrets
- made 'color' and 'posterImage" non-compulsory fields in Contentful (those two are need if a meetup is Featured, however making a meetup Featured is a human decision and whoever does that will need to update image and colour - this needs to be added to the relevant handover documentation for Marketing Team

## Review template


src/functions/meetup.js
```diff
@@ -128,15 +127,41 @@ const getSelfGroups = promisify(meetup.getSelfGroups.bind(meetup))
...
+// `console.log("content types are", collections.items);`
+//
+// Each content type has an ID that you can use as filter when calling `environment.getEntries({content_type: $contentTypeID})`
+// You can also find the content type ID on its Contentful page that lists its fields. This info is on the right hand side of the window (easy to miss on wider screens)
+//
+// space.getEntries() will be depreciated, use space -> environment -> entries
+
 exports.handler = async (event, context, callback) => {
   // Contentful user have many spaces. A space can have many environments.Each environment has entries of various "content models"
+  console.log("I'm about to run meetup function")
```
GBouffard: Remove

src/functions/meetup.js
```diff
@@ -146,16 +171,23 @@ exports.handler = async (event, context, callback) => {
       return null
     }
 
+    console.log(urlname, nextEvent)
```
GBouffard: remove

src/functions/meetup.js
```diff
@@ -146,16 +171,23 @@ exports.handler = async (event, context, callback) => {
...
+
     const meetup = processMeetupEvent(
       await getEvent({
         id: nextEvent,
         urlname
       })
     )
 
+    console.log(meetup)
```
GBouffard: remove

src/functions/meetup.js
```diff
@@ -146,16 +171,23 @@ exports.handler = async (event, context, callback) => {
...
+
     const ev = find(events, ['fields.linkToEvent.en-US', meetup.link])
     const entry = generateContentfulEvent({ ...meetup, ...group })
 
+    console.log(ev)
+    console.log(entry)
+
```
GBouffard: Remove both

src/functions/meetup.js
```diff
@@ -128,15 +127,41 @@ const getSelfGroups = promisify(meetup.getSelfGroups.bind(meetup))
...
   const space = await client.getSpace(CONTENTFUL_SPACE)
   const environment = await space.getEnvironment('master')
 
   // filter to return published entries that belong to a specific content model.
   const { items: events } = await environment.getEntries({
     limit: 1000,
     content_type: 'meetupEven'
+    // yes, the content type name is "meetupEven" - probably a typo during creation that can't be updated without recreating the contet from scratch
   })
```
GBouffard: You added a typo of your own in your comment ;)

README.md
```diff
@@ -20,8 +20,11 @@ CONTENTFUL_TOKEN=(see in contentful/settings/API keys "meetup > Content Delivery
 CONTENTFUL_SPACE=(see in contentful/settings/API keys "meetup > Space ID")
 MEETUP_KEY=(see in contentful/settings/API keys "gatsby > Content Delivery API - Access Token")
 GATSBY_ENVIRONMENT="development"
+CMS_CRUD=(copy from Netlify - Build & Deploy - Environment - Edit variables - CMS_CRUD. Read below why copy)
 ```
 
+The meetup contentful lambda script uses Contentful's Content Management API (https://www.contentful.com/developers/docs/references/content-management-api/). This requires a *Content Management Token* and NOT a Content Delivery Token (Content Delivery can only give you the info that's there, you can't CRUD entries). A CMT can only be viewed just after it has been generated. In theory these are Personal Tokens and each user can generate their own, but for simplicity's sake, it's easier for everyone to use the same one in their env file. Should you need to generate a new CMT, you can do so from contentful/settings/API keys/Content management tokens. The one used in Netlify is the "cms-crud" one.
```
omonk: you can link this with ```[Contentful's Content Management API](https://www.contentful.com/developers/docs/references/content-management-api/)```

omonk: You need to document here that the token used in production is a token registered with the apis@yld.io account on contentful.

src/functions/meetup.js
```diff
@@ -168,11 +192,16 @@ exports.handler = async (event, context, callback) => {
...
     const newEntry = await environment.getEntry(id.sys.id)
 
-    // console.log(`Publishing creted entry ${meetup.eventName}`)
+    console.log(`Publishing creted entry ${meetup.eventName}`)
     return newEntry.publish()
   })
+
+  return {
+    statusCode: 200,
+    body: 'Meetup function has finished running'
```
omonk: maybe its worth including what events were published to contenful here? It'll  be logged in the AWS lambda console and be helpful for debugging

omonk: ah ignore me this will be logged in the console logs above

webpack.functions.js
```diff
@@ -1,6 +1,8 @@
 const webpack = require('webpack')
+const nodeExternals = require('webpack-node-externals')
 
 module.exports = {
+  externals: [nodeExternals()],
```
omonk: what is this doing?

AniaMakes: The meetup function relies on several modules which are "backend" and throw errors when we try to bundle them. nodeExternals is the way to deal with this problem, per netlify-lambda (https://github.com/netlify/netlify-lambda#webpack-configuration)

AniaMakes:
> We don't want to delete `./functions/meetup.js`. look in `netlify.toml`, here we reference this lambda. without this file I assume netlify deployment will fail? or the lambda will just stop working?
> 
> I think it would be beneficial if you do some investigation into how these lambda functions are deployed via netlify. Granted I don't think that having this compiled function in the repo is useful or stable and should be deployed by something like serverless or something similar but if we don't have the knowledge into how lambdas these lambdas are deployed then we should leave it as is at the moment.

What does this comment refer to? I don't think I deleted `./functions/meetup.js`, have I?

omonk:
@AniaMakes ah indeed, the diff was so large it was hidden so appeared the file had been deleted

PR #200, opened 2019-05-13
Fix "when user double clicks on dropdown select button, text is highlighted"

AniaMakes:
## Description - [Trello ticket 462](https://trello.com/c/reHuM7mv/462-when-user-clicks-on-dropdown-select-button-text-is-highlighted-on-safari)

When user double clicked on the Navbar drop down (only items that have dropdown were affected), the word got selected per standard browser functionality. This is a patch to disable that selection. 

## Review template

### DIFF
--- a/src/components/Header/TopNav/Dropdown.js
+++ b/src/components/Header/TopNav/Dropdown.js
@@ -59,9 +59,6 @@ const DropdownNameWrapper = styled.span`
 const DropdownName = styled.span`
   padding-right: ${remcalc(6)};
   outline: none;
-  -webkit-user-select: none;
-  -moz-user-select: none;
-  -ms-user-select: none;
   user-select: none;
 `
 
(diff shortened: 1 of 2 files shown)


src/components/Header/TopNav/Dropdown.js
```diff
@@ -59,6 +59,10 @@ const DropdownNameWrapper = styled.span`
 const DropdownName = styled.span`
   padding-right: ${remcalc(6)};
   outline: none;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
```
omonk: styled components handles vendor prefixing, you should only need `user-select: none` here

AniaMakes: cunning!

PR #201, opened 2019-05-14
SideNav & TopNav upgrades + end-to-end testing

AniaMakes:
## Description - [Trello ticket 497-2](https://trello.com/c/9l0OD5Qr/497-top-nav-dropdown-refactoring)

Refactored the isExpanded logic so that it now works as ticket requires. 
Also added shortid module to solve "multiple children with same ID" warning in the console. 

- [x] SideNav dropdowns should remain expanded if the active page is within that drop down
- [x] TopNav - if the active page is in a dropdown, then the dropdown select button should be greyed out (i.e. use styles for "active" state)
- [x] e2e tests for Dropdown - for both TopNav (desktop) and SideNav (mobile)

## Review template


src/components/Header/SideNav/Dropdown.js
```diff
@@ -59,8 +60,17 @@ export default class Dropdown extends PureComponent {
   }
 
   render() {
-    const { items, children } = this.props
-    const { isExpanded } = this.state
+    const { items, children, path } = this.props
+    let { isExpanded } = this.state
+
+    const isServicePage = path.search(servicesRegExp) > -1
```
omonk: I think this is already written somewhere else. In utils?

src/components/Service/getSections.js
```diff
@@ -7,7 +9,7 @@ export default workStage =>
       // arrays start with 0 so that's why the + 1
       const getCurrent = part => `section${capitalize(part)}${index + 1}`
       return {
-        id: workStage.id,
+        id: `${workStage.id}-${generate()}`,
```
omonk: You don't need workStage at all here. Generate will suffice

src/components/Header/SideNav/Dropdown.js
```diff
@@ -59,8 +60,17 @@ export default class Dropdown extends PureComponent {
...
+    const serviceTitle = isServicePage ? path.match(servicesRegExp)[0] : null
+    const isSpecialityPage = path.includes('speciality')
+    const service = isSpecialityPage ? getService(path) : serviceTitle
+
+    isExpanded =
+      isExpanded ||
+      items.filter(item => item.to.startsWith(`/${service}`)).length > 0
```
omonk: Could this lead to not being able to close a dropdown?

AniaMakes: ugh, yes, good catch.

src/components/Service/getSections.js
AniaMakes: I wondered about that.

src/components/Header/SideNav/Dropdown.js
GBouffard: Good call. I added that 2 places for a previous ticket (TopNav components) but that wasn't part of a utils file. Now that it's used here as well it makes sense to move it to a utils file at the /Header level

GBouffard: I moved it 👍

src/components/Header/SideNav/Dropdown.js
omonk: Also I didn't notice this yesterday but you're referencing state and changing its value in the render without setting the state of the actual component which is not great. React has component lifecycle methods for what we're trying to do here

tests/e2e/sideNav-dropdown-test.js
```diff
@@ -0,0 +1,31 @@
...
+})
+
+test('a hamburger is present and opens the side nav', async t => {
+  const hamburger = await Selector('[class^="Hamburger"').filterVisible()
+  await t.expect(hamburger.exists).ok()
+  await t.click(hamburger)
+
+  const sideNav = await Selector('[class^="SideNav"]').filterVisible()
+  await t.expect(sideNav.exists).ok()
+})
```
omonk: So the main reason I wanted to get some tests for the sidenav was so we could catch any functionality errors. 

Lets add another one

`a dropdown opens when the dropdown title is clicked`

GBouffard: We have plenty of tests now:
 ✓ we are on the homepage
 ✓ a hamburger is present on the page
 ✓ the side nav is not expanded at its initial state
 ✓ when the hamburger button is clicked the side nav expands
 ✓ clicking a side Nav link redirects to the correct url and closes the side Nav
 ✓ when the side nav expands for the first time no sub-item is shown by default
 ✓ when an item is clicked its sub-items appear
 ✓ clicking a sub-item redirects to the correct url and closes the side Nav
 ✓ After a sub-item has been clicked and if the sideNav is re-opened, the subitems are still visible and the selected one highlighted

PS: We looked at testcafe-react-selectors but in the end, it was easier and more convenient to use the API made available by Testcafe.

tests/e2e/sideNav-dropdown-test.js
```diff
@@ -0,0 +1,129 @@
...
+
+  const joinUsLink = sideNavPanel.find('a').withText('Join us')
+  await t.click(joinUsLink)
+
+  const location = await getWindowLocation()
+  await t.expect(location.href).contains(`${baseUrl}/join-us`)
+  await t.expect(sideNavPanel.exists).notOk()
+})
+
+test('when the side nav expands for the first time no sub-item is shown by default', async t => {
```
omonk: Don't we auto expand dropdowns if the active page is within the dropdown?

omonk:
super thorough tests! nice one guys

tests/e2e/sideNav-dropdown-test.js
AniaMakes: Yes, but this example starts at `baseUrl` (line 34).

omonk: I was more referring to the name of the test. If anyone comes to the project should be able to garner a bit information on behaviour from tests. The test name should reflect that we don't open any dropdowns if none of the dropdowns contain links to the current path, we should also have a test to reflect the opposite

omonk:
rebased and merged in https://github.com/yldio/yld.io/pull/214

PR #202, opened 2019-05-14
Fix speciality events display

GBouffard:
## Description - Fix
The events query that was within pageQuery (Speciality) did not seem to work.
I updated so we retrieve all the events available (same query that we use on the homepage that Anna fixed and that we know works) and updated the logic to filter accordingly per speciality name + present & future dates.
It shows the actual 2 React events that are happening in the future / no other events shown for other specialities but they should show when we have some available.

## Review template


GBouffard:
Screen Shot 2019-05-14 at 11 52 48
How it looks with the React speciality

PR #206, opened 2019-05-20
Add scrollability to SideNav when height is not within window height

GBouffard:
## Description - [521](https://trello.com/c/Qqa85JQa/521-unable-to-scroll-sidenav-on-short-screens)

SideNav now scrollable on landscape views for Chrome.

## Review template

### DIFF
--- a/src/components/Header/SideNav/index.js
+++ b/src/components/Header/SideNav/index.js
@@ -16,7 +16,7 @@ const SideNavPanel = styled.nav`
   left: 0;
   top: 0;
   z-index: ${props => props.theme.zIndexes.header};
-  overflow-y: scroll;
+  overflow-y: auto;
   transform: translateX(100%);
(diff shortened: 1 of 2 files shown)


src/components/Header/SideNav/index.js
```diff
@@ -16,6 +16,7 @@ const SideNavPanel = styled.nav`
   left: 0;
   top: 0;
   z-index: ${props => props.theme.zIndexes.header};
+  overflow-y: scroll;
```
AniaMakes: Maybe `overflow-y: auto` might be better? 

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y says "overflow-y:scroll — always adds a scrollbar" and "overflow-y:auto — on most browser, equivalent to scroll"

GBouffard: - [ ] Done 👍 Still works.

PR #208, opened 2019-05-21
Chore : Contentful only updates when needed

AniaMakes:
## Description - [Trello ticket 519](https://trello.com/c/C97yFo4g/519-meetup-lambda-deploys-site-regardless-of-content)

Added logic so that if there is nothing new for a meetup that already exists in Contentful, there is no update. 

## Review template


functions/meetup.js
```diff
@@ -297,33 +296,85 @@
...
                   id: nextEvent,
                   urlname
                 })
               )
-              console.log(meetup)
               const ev = find(events, ['fields.linkToEvent.en-US', meetup.link])
               const entry = generateContentfulEvent({ ...meetup, ...group })
-              console.log(ev)
-              console.log(entry)
+              const presetHomepageFeatured = ev.fields.homepageFeatured
```
AniaMakes: Could I please have a sanity check that this will always remain the same rather even if ev is changed?

PR #210, opened 2019-05-21
Update noAfter to noafter (custom prop) and pass string value to remove warning

GBouffard:
## Description - [507](https://trello.com/c/aWCOWt9z/507-prevent-noafter-prop-from-being-added-to-anchor-tag)

- rename noAfter to noafter (no uppercase to custom prop)
- explicitly pass boolean as string value (also to avoid another warning on custom props)
- update snapshots

## Review template


src/components/Common/Statement.js
```diff
@@ -36,7 +36,7 @@ const Statement = ({ richText, children, noPadding }) => (
                   return (
                     <Link
                       key={content.data.uri}
-                      noAfter
+                      noafter="true"
```
AniaMakes: I don't know the codebase; I'm assuming that true is the default... What does it actually do and why would we use `false`?

omonk:
Can you investigate why this prop is making it to the anchor tag, that shouldn't be happening...

src/components/Common/Statement.js
omonk: So this prop is passed to the [styled-is](https://github.com/yldio/styled-is) library. (TIL it's made by yld 🎉)

I think you need to pass down a boolean here?

omonk: oh no you actually dont https://github.com/yldio/styled-is/blob/594f73dae6dc154cf884645ec53ffbdd49c0baaf/src/index.js#L12

jsms90: @AniaMakes, yes you are correct. If you pass in just a prop name (e.g. `noAfter`) without any particular value supplied to it, then yes, the value defaults to `true`. [That's true throughout React](https://stackoverflow.com/questions/37828543/how-to-pass-props-without-value-to-component), afaik

> What does it actually do

We can see that this `noAfter` prop is being applied to the `<Link>` component, which is our `StyledLink` under the hood. If you look at this file, you will see that the `noAfter` prop is being used here:
https://github.com/yldio/yld.io/blob/61f6e9bd1cef4882fbe90b4e7b016ebdb890fa34/src/components/Common/StyledLink.js#L20-L30

> why would we use false

Having a look at that bit of code (and by refreshing your knowledge of what `styled-is` library does for us, if you haven't already seen it), `noAfter="false"` would give you this CSS:
```
:after {
  content: none;
}
```

jsms90: Oh. And just in case this wasn't clear, the `:after` pseudo class is how our StyledLink component gets its underline.

For example, let's look at the 'Learn more' in the CaseStudyPreview component.

This is what it looks like without the `noAfter` prop:
image

and this is what it would look like *with* `noAfter`:
image


In my opinion, we shouldn't have this `noAfter` prop on the `StyledLink` component at all. And it has only been added as a bit of a hack, so that we can use `StyledLink` inside `Statement`, when actually it should be a different component under the hood.

GBouffard: ☝️ What Jen said. I din't update the logic. Just the syntax for react warning purposes.
`/Common/StyledLink` is used in a few places.
`Statement.js` is the only place where we used the `noafter` prop so the &:after style wont apply.
It applies everywhere else.
I would have rather passed the prop as `noafter` or `noafter={true}` but it gave me more warnings; suggested to pass `noafter="true"` so I did.

GBouffard: Agreed also with the comment that we hacked it to use `StyledLink` in here and it should have been done another way. This being said, refactoring or making a different component would not be high in term of priorities given all the new things we want to deliver.

jsms90: @GBouffard the `StyledLink` component is used in many places throughout our app. That's true. But all of the other locations are an appropriate usage of this component I.e. we want that functionality + styling.

This is the only situation (at least that I know of) where the usage of this `StyledLink` component is not inappropriate. Because it's not the right component to be using, we've ended up creating this `noAfter` prop, which is only being used within the context of `Statement`. If you search our codebase, there is no other situation where `noAfter` is being applied.

I would agree that this ticket is not super high priority, but since we're already completing the ticket & touching these files, we might as well fix the codebase, rather than just fixing the hack, no?

PR #211, opened 2019-05-22
Footer-Privacy-policy-and-cookies

GBouffard:
## Description - [206](https://trello.com/c/FuHmjNew/206-as-a-user-i-want-to-be-able-to-read-the-cookies-and-privacy-policies-even-after-having-clicked-agree-on-the-cookies-pop-up-foote)

- add links within Desktop and mobile/Tablet views.
Need feedback on design.

## Review template


omonk:
image
I believe we're missing the padding between the social links and the privacy policy links

GBouffard:
@omonk  The padding is on the GDPR links. There is already some padding on the social links and section as soon as it reaches a smaller device breakpoint. Also waiting for feedback from design team as there was no design in Zeppelin
Screen Shot 2019-05-23 at 09 36 54

src/components/Footer/index.js
```diff
@@ -37,62 +39,109 @@ export const OfficeStyled = styled(Row)`
...
+                    />
+                  )
+                })
+              }
+            </Locations>
+          </OfficeStyled>
+        </Padding>
+      </Grid>
+    </Padding>
+    <FollowUs>
```
jsms90: Given that this black banner now contains GDPR links, as well as social links, the name `FollowUs` is probably no longer appropriate for the black banner.

Perhaps you want to change this to `BlackBanner` or `BottomBanner` or `FooterEnd`, or something else that's better?

src/components/Footer/index.js
```diff
@@ -37,62 +39,109 @@ export const OfficeStyled = styled(Row)`
...
-        </Padding>
-      </FollowUs>
-    </GreyFooter>
-  )
-}
+    </FollowUs>
+    <Grid>
+      <Row>
+        <Col width={1}>
+          <Padding bottom={3}>
```
jsms90: I think we agreed that we weren't going to use this component any more & that we would slowly remove `styled-components-spacing`'s components from the codebase whenever we touched an old file.

src/components/Footer/index.js
```diff
@@ -37,62 +39,109 @@ export const OfficeStyled = styled(Row)`
...
+  ${breakpoint('desktop')`
+    visibility: hidden;
+  `}
+`
+
+const LinkUnderline = styled(Link)`
+  text-decoration: underline;
+`
+
+const gdprLinksElement = gdpr.map(link => (
```
jsms90: It's probably more declarative/readable to have this map sitting inside the JSX below. We don't need to extract it out here, since we're not re-using it anywhere.

src/components/Footer/index.js
```diff
@@ -37,62 +39,109 @@ export const OfficeStyled = styled(Row)`
...
+  a {
+    margin-left: 24px;
+  }
+
+  ${breakpoint('desktop')`
+    visibility: visible;
+  `}
+`
+
+const MobileGDPR = styled.div`
```
jsms90: I don't think we need to create 2 different components - 1 for desktop & 1 for mobile - since the content and functionality is exactly the same. Layout is better generally dealt with via CSS than creating 2 sets of HTML, if you can help it.

In the case of our project, I think the way that we generally handle this kind of situation is through the `Col` component.

If you wrap `<Social>` in one `Col` & `<GDPRLinks>` in another `Col`, then you can use the `width` prop on `Col` to control their width on different screen sizes.

GBouffard: Yes . I also considered that option and something was not sitting right with me by doing 2 components. It makes sense to control via CSS.
I have done 1 component now and so far the mobile/tablet view.
Now I have a problem; I don't understand how to make turn a `Col` to 50% width. Can't find much info on it in the repo either. I see sometimes that we do stuff like `<Col width={[1, 1, 1, 1, 3 / 12]}>` but that's not speaking to me...

src/components/Footer/index.js
GBouffard: Good idea. I mixed your solution and it's now `FooterBanner`

src/components/Footer/index.js
```diff
@@ -37,62 +39,109 @@ export const OfficeStyled = styled(Row)`
...
-              </Col>
-            </Row>
-          </Grid>
-        </Padding>
-      </FollowUs>
-    </GreyFooter>
-  )
-}
+    </FollowUs>
+    <Grid>
```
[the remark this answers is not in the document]
GBouffard: Now, we have 1 component inside `FooterBanner` (previously known as `FollowUs` so it should not happen anymore). Thanks for the explanation of what happens though.

src/components/Footer/index.js
GBouffard: Not using anymore in my code as using 1 component but I didn't know about the `styled-components-spacing` not being used anymore. I found that the way we use CSS and the settings we have (see my comment on `Col` and `Row`) would need more documentation.

src/components/Footer/index.js
jsms90: I agree that it would be useful for us to document how to use our grid system but [the ticket for creating docs](https://trello.com/c/DrXHIMWe/31-document-how-our-grid-works-should-be-used-in-combination-with-the-other-style-libraries-we-use) has been deprioritised.

Our project's grid system is built on a package called `rebass grid` - [here's the section on responsive styles](https://github.com/rebassjs/grid#responsive-styles). Rebass grid is built on [`styled system`](https://github.com/styled-system/styled-system), which has [slightly more detailed docs](https://styled-system.com/responsive-styles) on this `width` prop.

So to summarise - you achieve 50% width by writing `width={1/2}`.
But we have an array, because this corresponds to the array of breakpoints in our `theme.js`. So the first four `1`s in the array correspond to the mobile breakpoints & the `3/12` will correspond to small tablet and upwards. If we wanted large tablet to have a different width, we would add another number to the array. And so on.

Hope that makes slightly more sense?

src/components/Footer/index.js
jsms90: I believe that we agreed this on a call between you me and David, that we would slowly remove `Padding` components from the codebase. I thought you seemed as excited as me :sweat_smile:

src/components/Footer/index.js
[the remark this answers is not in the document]
jsms90: Looks good :+1:

src/components/Footer/index.js
```diff
@@ -37,62 +39,84 @@ export const OfficeStyled = styled(Row)`
...
+  a {
+    padding-right: 36px;
+  }
+`
+
+const LinkUnderline = styled(Link)`
+  text-decoration: underline;
+`
+
+const gdprLinksElement = gdpr.map(link => (
```
jsms90: Now that this isn't repeated anywhere, we can probably move this code directly into line 113 so that all the code is together

But it still might be nice to separate the Office section & FooterBanner sections into different files, just so that this file isn't so long?

src/components/Footer/index.js
GBouffard: Haha! I don't like these `Row` `Col` `Padding` components as I don't find them explicit enough but I don't remember what we agreed. I feel like Dory from Finding Nemo right now; short-term memory 😆

src/components/Footer/index.js
GBouffard: It did! Nice and updated 👍 Thanks

src/components/Footer/index.js
GBouffard: Absolutely! 👍 Done

src/components/Footer/index.js
jsms90: Yeah I don't blame you :sweat_smile:

PR #212, opened 2019-05-22
Fix www.yld.io/contact/ form textarea being resizable beyond its parent.

Skeletonxf:
On the https://www.yld.io/contact/ form the textarea with an id of `message` can be resized by the user as textareas can be by default. This means it is possible to extend the width of the textarea beyond that of its parent by accident.

Screenshot_2019-05-22 YLD - Contact - Software engineering, web design, training and open source consultancy YLD(1)

Screenshot_2019-05-22 YLD - Contact - Software engineering, web design, training and open source consultancy YLD

Once this happens it is not so easy to resize it back down, as the corner to drag to resize the textarea becomes cut off.

Add `max-width: 100%;` to the Input component which prevents the textarea in the form from being resized beyond its parent width but still allows for users to resize it vertically or shrink the width.

## Review template


omonk:
@Skeletonxf awesome thank you for your help!

PR #215, opened 2019-05-24
add locale to contentful entry object

omonk:
Noticed we weren't giving the contentful entry a local 'en-US' so contentful would probably reject the update or just ignore it


omonk:
I didn't read the code hard enough. Thanks @AniaMakes

PR #217, opened 2019-05-28
Why Open Source component (v2)

AniaMakes:
## Description - [Trello ticket 484](https://trello.com/c/6bf4cfEm/484-why-use-open-source-component)

Created Why-Open-Source component, as well as a storybook story to showcase it.

## Review template


src/components/OpenSource/WhyOpenSource.js
```diff
@@ -0,0 +1,27 @@
...
+
+const StyledSubtitleWithBody = styled(SubtitleWithBody)`
+  padding-top: ${props => props.theme.spacing['1']};
+  padding-bottom: ${props => props.theme.spacing['1']};
+  margin-top: ${props => props.theme.spacing['1']};
+  margin-bottom: ${props => props.theme.spacing['2']};
+`
+
+const StyledImage = styled.img`
+  padding-top: ${props => props.theme.spacing['2']};
```
jsms90: I'm assuming that you have added `24px` padding above the image:
image

so that there will be a separation between different `LogoSubtitleBody`s?

But don't forget that this means the first `LogoSubtitleBody` will also have `24px` padding above it, which we don't want because we want the top of this component to line up with the title of this section:
image

This `24px` padding is probably best applied within the overall section component. Think [Atomic Design](http://atomicdesign.bradfrost.com/)

src/components/OpenSource/WhyOpenSource.js
```diff
@@ -0,0 +1,27 @@
+import React, { Fragment } from 'react'
+import SubtitleWithBody from '../Common/SubtitleWithBody'
+import styled from 'styled-components'
+
+const StyledSubtitleWithBody = styled(SubtitleWithBody)`
```
jsms90: I believe the padding is already correct on `SubtitleWithBody` & if you remove these lines, then the component in storybook will look the same anyway.

Besides, why do we need extra spacing on the `SubtitleWithBody` component?

image

stories/LogoSubtitleBody.stories.js
```diff
@@ -0,0 +1,30 @@
...
+
+import icon from './assets/community_support.svg'
+
+addDecorator(Theme)
+
+const title = 'Community support'
+const body =
+  'Open source technologies are supported by knowledgeable and resourceful communities. Their collective contributions result in more robust, innovative and faster solutions.'
+
+storiesOf('Logo, Subtitle & Body', module).add('on white Background', () => {
```
jsms90: Could we have the name of this story reflect the name of the React component? :pray: 
i.e. `LogoSubtitleBody` rather than comma separated

That way, someone (e.g. a designer) just looking at storybook already knows exactly what the component is called in our codebase

src/components/OpenSource/WhyOpenSource.js
```diff
@@ -0,0 +1,27 @@
...
+  padding-top: ${props => props.theme.spacing['1']};
+  padding-bottom: ${props => props.theme.spacing['1']};
+  margin-top: ${props => props.theme.spacing['1']};
+  margin-bottom: ${props => props.theme.spacing['2']};
+`
+
+const StyledImage = styled.img`
```
jsms90: We want to use [our `Image` component](https://github.com/yldio/yld.io/blob/master/src/components/Common/Image.js) under the hood, rather than just `img` :pray: 

If you haven't seen it before, our `Image` component is used to ensure that we're playing nicely with the [`gatsby-image` plugin](https://www.gatsbyjs.org/docs/using-gatsby-image/) - a plugin which optimises images for us & helps with responsiveness, etc.

For example, this line in `Image.js`:
https://github.com/yldio/yld.io/blob/16e4e6b4034811823412e49d16adee9b7c23e353/src/components/Common/Image.js#L6

refers to the `fluid` prop that we get from the plugin. The fluid prop helps us with responsivity - [see the docs](https://www.gatsbyjs.org/packages/gatsby-image/#two-types-of-responsive-images).

You probably don't need to worry about understanding all of this right now. But at the very least, it's useful to know that all of this exists :smiley:

stories/LogoSubtitleBody.stories.js
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Grid>
+      <Row>
+        <Col width={[1, 1, 1, 1, 0.5]}>
```
jsms90: I would leave `Grid`, `Row` & `Col` out of the `LogoSubtitleBody` story because this small component may be re-used in other contexts where there are a different number of columns.

You would need `Grid`, `Row` & `Col` for the overall WhyOpenSource section though :+1:

stories/LogoSubtitleBody.stories.js
```diff
@@ -0,0 +1,30 @@
...
+          <WhyOpenSource
+            icon={icon}
```
jsms90: Maybe choose either `icon` or `logo`, in terms of naming? So either call this component `IconSubtitleBody` or rename this prop `logo`?

PR #218, opened 2019-05-28
Update copy and remove placeholder in contact form

GBouffard:
## Description - [Updates on 510](https://trello.com/c/qkY0xr5X/510-corrections-to-form-fields-on-contact-us)

(see comments from James)
- [ ] update copy
- [ ] Remove placeholder completely

## Review template

### DIFF
--- a/src/pages/contact.js
+++ b/src/pages/contact.js
@@ -73,12 +73,31 @@ class ContactUs extends Component {
     const { name, email, message, submitting, success } = this.state
     const {
       location,
-      data: { contentfulPage: page }
+      data: {
+        contentfulPage: {
+          title,
+          slug,
+          seoTitle,
+          seoMetaDescription,
+          titleNotContacted,
+          titleContacted,
+          labelInterests,
+          interests,
+          labelYourMessage,
+          labelYourName,
+          labelYourEmail,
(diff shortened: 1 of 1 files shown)


src/pages/contact.js
```diff
@@ -29,12 +29,11 @@ const data = {
   },
   labels: {
     interests: 'What are you interested in?',
-    yourName: 'Your Name',
+    yourName: 'Your full name',
```
AniaMakes: I know we probably don't have much say onto what "yourName" text is, but for inclusion reasons I'd argue to change it back to "Your Name". We have no business asking for people's middle names (and full name implies that). Full name also implies "needs to match legal document" and then you run into issues with people who use different name than the one they have in their documents. Also: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

jsms90: @GBouffard has implemented what's been asked for, so code-review wise I think we're good.

@AniaMakes if you feel strongly about there being different copy from what has been requested, it's probably something to bring up with product/design?

GBouffard: @AniaMakes This is now editable directly in Contentful so Marketing manages these matters.

PR #219, opened 2019-05-29
add no-unused-vars, fix code

omonk:
Codebase had a bunch of unused vars everywhere, this PR changes this and adds an eslint rule to make sure it doesn't happen

### DIFF
--- a/src/components/Footer/Office.js
+++ b/src/components/Footer/Office.js
@@ -13,7 +13,7 @@ const Office = ({ name, telephone, email, streetAddress }) => (
   <PaddedCol width={[1, 1, 1, 1 / 2, 1 / 2, 1 / 4]}>
     <Subtitle reverse>{name}</Subtitle>
     <Padding top={1} bottom={1}>
-      {streetAddress.map((address, i) => (
+      {streetAddress.map(address => (
         <StyledBodyPrimary key={address}>{address}</StyledBodyPrimary>
       ))}
       <StyledBodyPrimary itemProp="telephone">{telephone}</StyledBodyPrimary>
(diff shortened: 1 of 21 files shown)


src/components/Footer/Office.js
```diff
@@ -13,7 +13,7 @@ const Office = ({ name, telephone, email, streetAddress }) => (
   <PaddedCol width={[1, 1, 1, 1 / 2, 1 / 2, 1 / 4]}>
     <Subtitle reverse>{name}</Subtitle>
     <Padding top={1} bottom={1}>
-      {streetAddress.map((address, i) => (
+      {streetAddress.map(address => (
         <StyledBodyPrimary key={address}>{address}</StyledBodyPrimary>
```
jsms90: Better to change this to `key={i}`?

src/functions/meetup.js
```diff
@@ -260,7 +261,6 @@ exports.handler = async (event, context, callback) => {
       console.log(
         `Not prod so not creating contentful event for ${meetup.eventName}`
       )
-      return
```
jsms90: is this not needed?

src/components/Footer/Office.js
omonk: Using a map index arg as a key isn't a great pattern

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

omonk: address is unique enough here and its not dynamic so it'll be ok

src/functions/meetup.js
omonk: Not really, we're not using the return value from this iteration

src/components/Footer/Office.js
jsms90: kk

PR #224, opened 2019-05-30
Why Open Source Component 

AniaMakes:
## Description - [Trello ticket 484](https://trello.com/c/6bf4cfEm/484-why-use-open-source-component)

- changed `titleAndList` component to `titleAndListWithOptionalLogos` and added a story with logos
- created a `WhyOpenSource` component that has `titleAndListWithOptionalLogos`, `Subtitle`, and `LogoGrid`
- created a story to go with `WhyOpenSource`

## Review template


src/components/OpenSource/WhyOpenSource.js
```diff
@@ -0,0 +1,22 @@
+import React, { Fragment } from 'react'
+import { Padding } from 'styled-components-spacing'
+
+import TitleAndListWithOptionalLogos from '../Common/TitleAndListWithOptionalLogos'
+import Subtitle from '../Typography/Subtitle'
+import LogoGrid from '../Common/LogoGrid'
+
+const WhyOpenSource = ({ title, list, subtitle, extraContent, companies }) => {
+  return (
+    <Padding>
```
omonk: We're not using `Padding` component anymore

Use `styled-components` and create your own component using the `props.theme.spacing[n]`

src/components/Common/TitleAndListWithOptionalLogos.js
```diff
@@ -16,6 +17,17 @@ const ItemBody = styled(Body)`
     padding: 0;
   }
 `
+
+const SquareImg60 = styled(Image)`
+  max-width: 60px;
```
GBouffard: We usually like to use remcalc
`import remcalc from 'remcalc'`
and then you can update all your `XXpx` to `${remcalc(XX)}` instead.

stories/titleAndListWithOptionalLogos.stories.js
```diff
@@ -4,16 +4,23 @@ import Theme from './theme'
...
 import StyledLink from '../src/components/Common/StyledLink'
 
 addDecorator(Theme)
 
+const ExampleImage = {
+  title: 'Community Support Logo',
+  file: { url: '/community_support.svg' },
+  className: 'logo'
+}
+
```
GBouffard: This is now used twice ☝️  Maybe worth making an export import in one of the files.
Same with body, or any re-used filed.

src/components/Common/TitleAndListWithOptionalLogos.js
omonk: I believe its ok to use hardcoded values here. things like padding etc. should be remcalc etc.

omonk: wait sorry @GBouffard is totally right, ignore me 🤦‍♂

jsms90:
@AniaMakes This is what the story looks like on storybook. Seems like something is going wrong with the images?
image

stories/titleAndListWithOptionalLogos.stories.js
jsms90: I don't think we will need the WhyOpenSource story, once the section is on the open source page. So it's prob fine to leave the duplication for now?

stories/titleAndListWithOptionalLogos.stories.js
```diff
@@ -4,16 +4,23 @@ import Theme from './theme'
...
 import { Grid } from '../src/components/grid'
-import TitleAndList from '../src/components/Common/TitleAndList'
+import TitleAndList from '../src/components/Common/TitleAndListWithOptionalLogos'
 import StyledLink from '../src/components/Common/StyledLink'
 
 addDecorator(Theme)
 
+const ExampleImage = {
+  title: 'Community Support Logo',
+  file: { url: '/community_support.svg' },
```
jsms90: './assets/community_support.svg' ?

It might be nice to import at the top of the file & then pass in here like:
`import CommunitySupportLogo from './assets/community_support.svg'`
then
```
const ExampleImage = {
  title: 'Community Support Logo',
  file: { url: CommunitySupportLogo }
}
```

Does that work?

src/components/Common/TitleAndListWithOptionalLogos.js
AniaMakes: Going to leave hardcoded values for now.

stories/titleAndListWithOptionalLogos.stories.js
AniaMakes: I don't think that would work because the way storybook works, is that you give it a path on build (it was missing in package.json, that's why your screenshot above doesn't have them), and the path is `./stories/assets`. Storybook knows to automatically look into that folder when given a link.

jsms90: Nah. My misunderstanding.

Thanks!

I'm not sure why your images are broken on storybook then?

PR #234, opened 2019-05-31
Open source talks section

GBouffard:
## Description - [528](https://trello.com/c/04vFF1sg/528-talks-section-for-os-page)

- [x] Move 3 columns of talks pattern (name: VideoGrid ?) out of JoinUs page sub-section & into Common folder as its own React component
- [x] Compose icon, title, new VideoGrid component & StyledLink into TalksSection within OpenSource folder
- [x] Edit graphql query on open-source.js, to pull content from Contentful, & use Talks Section on Open Source page

## Review template


src/components/OpenSource/Partnerships.js
```diff
@@ -9,7 +10,15 @@ import Image from '../Common/Image'
 import SubtitleWithBody from '../Common/SubtitleWithBody'
 
 const StyledSectionTitle = styled(Col)`
```
jsms90: Why do you need all of these paddings on top of the `Col` component's `width` prop? The spacing should be handled by the grid :confused:

src/components/OpenSource/Talks.js
```diff
@@ -0,0 +1,74 @@
...
+import CompactVideoLink from '../Common/CompactVideoLink'
+import StyledLink from '../Common/StyledLink'
+import { Grid, Row, Col } from '../grid'
+import { SectionTitle } from '../Typography'
+
+const StyledImage = styled.img`
+  padding: ${remcalc(83)} 0 ${remcalc(15)};
+`
+
+const StyledSectionTitle = styled(Col)`
```
jsms90: See top comment

src/components/OpenSource/Talks.js
```diff
@@ -0,0 +1,74 @@
...
+import remcalc from 'remcalc'
+
+import BlueBackground from '../Common/BlueBackground'
+import CompactVideoLink from '../Common/CompactVideoLink'
+import StyledLink from '../Common/StyledLink'
+import { Grid, Row, Col } from '../grid'
+import { SectionTitle } from '../Typography'
+
+const StyledImage = styled.img`
+  padding: ${remcalc(83)} 0 ${remcalc(15)};
```
jsms90: Where are these values coming from?

You need `72px` above and below the whole component (i.e. applied to `Grid`) - please use the `space` array in our theme & don't forget that other breakpoints will have different values.

The only other padding you might want to apply to the image is in order to achieve the `24px` between the image and the title below it:

image

src/components/OpenSource/Talks.js
```diff
@@ -0,0 +1,74 @@
...
+
+const StyledTalksLink = styled(StyledLink)`
+  padding-bottom: ${remcalc(90)};
+`
+
+const TalkCol = ({
+  title,
+  type,
+  link,
+  cta,
```
jsms90: You aren't getting all of these fields from Contentful & there's no need to destructure props that you don't use anyways.

src/pages/open-source.js
```diff
@@ -38,6 +52,20 @@ const OpenSourcePage = props => (
...
+          talksSectionImage {
+            title
+            file {
+              url
+            }
+          }
+          talksSectionTitle
+          talksSectionTalks {
+            title
+            type
```
jsms90: We don't need this field. The type should always be 'talk'.

_But_, since we can't force Marketing to enter `ExternalResource`s which are only of type 'talk', we _could_ use the type field to make a check here in our codebase. Since you're pulling the type out here anyway, it might be an idea to do this check?

src/components/OpenSource/Talks.js
```diff
@@ -0,0 +1,74 @@
...
+  <BlueBackground>
+    <Grid>
+      <StyledImage src={icon.file.url} alt={icon.title} />
+      <Row>
+        <StyledSectionTitle width={[1, 1, 1, 1, 1 / 2, 1 / 2, 1 / 2]}>
+          <SectionTitle reverse>{title}</SectionTitle>
+        </StyledSectionTitle>
+      </Row>
+      <StyledTalksRow>
+        {talks && talks.length && talks.map(talk => TalkCol(talk))}
```
jsms90: Can you put this `talks &&....` defensive coding around the whole `TalksSection` on the `open-source.js` file? If there are no talks at all coming from Contentful, we don't want to render any of this section at all. This code would leave the section title & CTA and simply render no videos, which is not quite what we want

src/pages/open-source.js
```diff
@@ -12,6 +13,11 @@ const OpenSource = ({ data }) => {
       slug,
       seoTitle,
       seoMetaDescription,
+      talksSectionImage,
```
jsms90: We could group these as @omonk is doing in his PR & then pass in `{...data}` to the `TalksSection`. What do you think @GBouffard @omonk ?

For example, the contributions section in `open-source.js`:
https://github.com/yldio/yld.io/blob/af1490401b5ef88e7bcfe7256987e2637945e412/src/pages/open-source.js#L12-L29

and then destructuring within `Contributions.js`:
https://github.com/yldio/yld.io/blob/af1490401b5ef88e7bcfe7256987e2637945e412/src/components/OpenSource/Contributions.js#L28-L48

src/pages/open-source.js
```diff
@@ -21,6 +27,14 @@ const OpenSource = ({ data }) => {
...
       <Head page={{ title, slug, seoTitle, seoMetaDescription }} />
       {/* <CaseStudyPreview caseStudy={} /> */}
+      <TalksSection
+        icon={talksSectionImage}
+        title={talksSectionTitle}
+        talks={talksSectionTalks}
+        ctaText={talkSectionCtaText}
+        ctaLink={talksSectionCtaLink}
+      />
+      <p>open source page</p>
```
jsms90: Please rebase onto the latest `open-source-page` branch, to avoid putting things like this back in :pray:

src/components/OpenSource/Talks.js
```diff
@@ -0,0 +1,74 @@
...
+const TalkCol = ({
+  title,
+  type,
+  link,
+  cta,
+  featured,
+  additionalInfo,
+  colorCode
+}) => (
+  <CompactVideoLink
```
jsms90: What controls the outline colour?

Zeplin
image

Your branch
image

src/components/OpenSource/Talks.js
```diff
@@ -0,0 +1,74 @@
...
+    {title}
+  </CompactVideoLink>
+)
+
+const TalksSection = ({ icon, title, talks, ctaText, ctaLink }) => (
+  <BlueBackground>
+    <Grid>
+      <StyledImage src={icon.file.url} alt={icon.title} />
+      <Row>
+        <StyledSectionTitle width={[1, 1, 1, 1, 1 / 2, 1 / 2, 1 / 2]}>
```
jsms90: You're not styling a `SectionTItle`, you're styling a `Col` here. A more accurate name for this component would be `StyledCol` :pray: 

Right now your code reads: "I have a SectionTitle inside a StyledSectionTitle"

src/components/OpenSource/Talks.js
GBouffard: YES!!! Awesome idea, the parent should decide. Also I did it for partners as I used a similar logic. Great stuff Jen! Done.

src/pages/open-source.js
GBouffard: Done.

src/components/OpenSource/Talks.js
GBouffard: const MAX_VIDEOS is a good idea. Used already existing code which is pretty much doing what I was doing. Still need to resolve the outline color.

src/pages/open-source.js
GBouffard: True. I was also thinking about that and then forgot about it. Since we only have 1 type at the moment it's not really necessary I'd say. As soon as we would get at least 2, we could. What do you think?

src/pages/open-source.js
GBouffard: By passing `...data` you also pass a lot of things that the component doesn't need. I'd be more in favour of destructuring on top and passing what is needed to each. It sounds like it'd also be more efficient. What do you guys think?

src/components/OpenSource/Talks.js
GBouffard: Yeah, pretty much. I wished I knew . that before; would have saved some time to all of us.

src/pages/open-source.js
jsms90: When you use the `ExternalResource` content type, in Contentful, this is the view:
image

Here is one where the `type` is talk:
image

But there are others where the `type` is something different:
image

Here is the place in Contentful where we attach the "Talks" to the open source page - this requires Marketing to enter an _External Resource_ but we have no way of restricting Marketing to only put in External Resources _of type: 'talk'_.

So, knowing that someone could accidentally put some other type of external resource in here, I reckon we should check in the codebase & strip out anything we're getting from Contentful which is not of type: 'talk'.

Do you see what I mean?

GBouffard:
BTW, I merged the latest version of master in the branch as well. If it has not been done with `open-source-page` it explains why the diff is so big

src/pages/open-source.js
jsms90: Yeah I ended up saying exactly the same thing actually, on my PR: https://github.com/yldio/yld.io/pull/237#issuecomment-498209125

I guess the thing I was pointing out here was about us having a consistent approach. @omonk any thoughts?

src/components/OpenSource/Talks.js
jsms90: On the Join Us page, the outline is already grey:
image

Perhaps look into how that was achieved?

jsms90: Hmm

Ok...judging by [Zeplin designs for `StandaloneVideoLink`](https://app.zeplin.io/project/5b4e666f0d4230a16dffd93a/screen/5c3e0570797c210e8b18b0d3):
image

the border, in line 77 of this code should be changed based on `themeVariation`:
https://github.com/yldio/yld.io/blob/5f40328f15a59d285b02363cae2b9668a92132bc/src/components/Common/BaseVideoLink.js#L72-L78

src/pages/open-source.js
GBouffard: Oh. I see. Ok, in that case I will add a safety net so we only get `talks`. Thanks for the detailed explanation and screenshots

src/components/OpenSource/Talks.js
GBouffard: ok, so it's something that I can fix but it will fix for every video using the dark themevariation.

src/pages/open-source.js
omonk: It's more of a preference to be honest. passing props as `{...data}` doesn't have any difference in the resolution of data in JS. My reason for this approach is to localise any logic for a component to that component rather than having `open-source.js` contain a bunch of logic for all of the different sections/components. 

It's a static site so there is little business logic that we need to resolve as its mostly quite flat data.

src/components/OpenSource/Talks.js
GBouffard: Done. Thanks Jen!

src/pages/open-source.js
GBouffard: Cool. Thanks for the feedback Ollie. Let's park this for now then.

src/components/OpenSource/Partnerships.js
GBouffard: Done with the Grid. Thanks Jen

src/components/OpenSource/Talks.js
GBouffard: Done with the Grid. Thanks Jen

src/components/OpenSource/Talks.js
GBouffard: Done. I updated StandaloneVideoLink after I spoke with James.

src/pages/open-source.js
```diff
@@ -21,23 +20,40 @@ const OpenSource = ({ data }) => {
...
   } = data
 
+  const talks = talksSectionTalks.filter(({ type }) => type === 'Talk')
+
   return (
     <Layout>
       <Head page={{ title, seoTitle, seoMetaDescription }} />
       <CaseStudyPreview isTop caseStudy={featuredCaseStudy} />
       <Statement>{statement}</Statement>
-      <OpenDeliverables {...data} />
```
omonk: We need this in here

src/pages/open-source.js
```diff
@@ -4,14 +4,13 @@ import { StaticQuery, graphql } from 'gatsby'
...
 import BlueBackground from '../components/Common/BlueBackground'
-import { Grid } from '../components/grid'
 import CaseStudyPreview from '../components/Common/CaseStudyCards/CaseStudyPreview'
 import Statement from '../components/Common/Statement'
 
+import TalksSection from '../components/OpenSource/Talks'
 import PartnershipsSection from '../components/OpenSource/Partnerships'
 
 import Contributions from '../components/OpenSource/Contributions'
-import OpenDeliverables from '../components/OpenSource/OpenDeliverables'
```
omonk: this as well

src/pages/open-source.js
GBouffard: Oh. Makes sense. I imported it and it was telling me `OpenDeliverables` is not used anywhere so I deleted 😆  I will put back

src/pages/open-source.js
GBouffard: Yep, related to the above which I probably removed by mistake.

src/pages/open-source.js
omonk: :D

PR #237, opened 2019-05-31
add whyOS component to open source page

jsms90:
## Add why os to open source page - [Trello ticket](https://trello.com/c/6bf4cfEm/484-why-use-open-source-component)

- add `WhyOpenSource` component that was created via PR https://github.com/yldio/yld.io/pull/224 to `open-source.js`

To do:
- apply paddings around component, based on each breakpoint
- may need to refactor (e.g. alter the spacing and/or application of `Row`, `Col`, etc) that's applied individually to sub-components


jsms90:
Yeah it's true. I could destructure all the data we get back from Contentful about this page & pass the entire data set into `WhyOpenSource`. But in general, isn't it preferable to be more declarative than that? https://stackoverflow.com/questions/52621169/react-props-should-i-pass-the-object-or-its-properties-does-it-make-much-diffe

I feel like it's entirely appropriate for us to be manipulating the data that we get from Contentful into the right shape, at the page level where we're doing the querying. Then we can just plug the correct data into our super functional little sub sections, rather than tightly coupling the two.

PR #242, opened 2019-06-04
contact us page updates

omonk:
A quick requirement to include contact us in the top nav, footer links and adding the `GetInTouch` component to the `Footer` component


src/components/Footer/index.js
```diff
@@ -1,32 +1,54 @@
...
-        <OfficeListing />
-      </Grid>
-    </Padding>
-    <FooterLinks />
-  </GreyFooter>
+  <Fragment>
+    <Location>
+      {({ location: { pathname } }) =>
+        showGetInTouch(pathname) ? (
+          <GetInTouch
```
GBouffard: Given the urgency of the ticket I wouldn't focus on making the change but any string like that should probably come from Contentful.

src/components/Footer/index.js
```diff
@@ -1,32 +1,54 @@
...
 import OfficeListing from './OfficeListing'
 import FooterLinks from './FooterLinks'
 
 export const GreyFooter = styled.footer`
   background: #232323;
 `
 
+const showGetInTouch = pathname => {
+  const path = pathname.replace(/^\/|\/$/gi, '')
+  return !['contact', 'training', 'speciality/node-js', 'join-us'].includes(
```
GBouffard: I would have the array in `src/components/Header/navLinksHelper.js`

src/components/Footer/FooterLinks.js
```diff
@@ -58,34 +58,36 @@ const LinkUnderline = styled(Link)`
...
-                  {link.label}
-                </LinkUnderline>
-              ))}
-            </GDPR>
-          </Col>
-        </Row>
-      </Grid>
-    </Padding>
-  </FooterBanner>
+  <Fragment>
```
GBouffard: Fragment not needed

src/components/Common/GetInTouch.js
```diff
@@ -13,8 +12,8 @@ const RowLayout = styled(Row)`
   `}
 `
 
-const GetInTouch = ({ title, contactText }) => (
-  <GreyBackground>
+const GetInTouch = ({ title, contactText, ctaText = 'Get in touch' }) => (
+  <div>
```
GBouffard: I dont understand. I see `Get in touch` already in the contact page. has something been deleted from contentful? If that's the case the solution is to make the field mandatory in Contentful

AniaMakes:
image

There are places where you have many white background components in the row. Is that OK with the design team?

src/components/Footer/index.js
omonk: You're right. I want to make a Meta content type that will deal with things like this and the top nav and footer links etc.

src/components/Footer/index.js
omonk: I'd rather keep it in here. It's localised to the footer and doesn't need to be reused elsewhere

omonk:
> image
> 
> There are places where you have many white background components in the row. Is that OK with the design team?

@AniaMakes - yep I went over this with Hana. Also I've removed the footer contact us from any specialities page so this wont be here anymore

src/components/Footer/links.js
```diff
@@ -51,6 +51,10 @@ const social = [
 ]
 
 const gdpr = [
+  {
+    label: 'Contact Us',
```
GBouffard: should it be Contact `u`s ? It's `Join us` at the moment on the website

src/components/Footer/index.js
GBouffard: 👍

src/components/Footer/index.js
GBouffard: No worries. If not used elsewhere it's valid to leave it here.

src/components/Footer/index.js
```diff
@@ -1,32 +1,76 @@
...
 import FooterLinks from './FooterLinks'
 
 export const GreyFooter = styled.footer`
   background: #232323;
 `
 
+const Wrapper = styled.div`
+  background-color: ${({ theme, bgColor = 'greyBg' }) =>
+    theme.colors[bgColor || 'white']};
+`
```
GBouffard: Since bgColor has a default value,  I believe that the 'white' should never be reached, right? Unless bgColor has a value that we don't have in theme.color; is that the reason we have the fallback to white?
PS: the destructuring of props is nice! I realise that we hardly do that in our styling. We should whenever possible.

omonk: You're absolutely right thanks for spotting that

GBouffard: you can remove the `|| 'white'`. Only `=> theme.colors[bgColor]` needed now 😉

GBouffard: Oh sorry...that's what you did. Perfect. I was seeing an older version 😄

PR #263, opened 2019-06-17
Seo updates

omonk:
This PR aims to solve 3 main issues currently on the site 
- Missing h1 tags on any page, specialities and services pages were missing.
- Adds `react-media`, this package means we prevent duplicate content in `h1` tags specifically the case study preview section. 
- Updates all of the anchor tags PropTypes to ensure we have a valid title prop for each anchor

### DIFF
--- a/src/components/Common/CaseStudyCards/CaseStudyHero.js
+++ b/src/components/Common/CaseStudyCards/CaseStudyHero.js
@@ -15,7 +15,7 @@ const FlexCol = styled(Col)`
 const SubSection = ({ heading, items }) => (
   <Padding bottom={0.5}>
     <Subtitle noPaddingBottom>{heading}</Subtitle>
-    <Flex alignCenter wrap>
+    <Flex alignCenter wrap="true">
       <SeoLinks noPadding items={items} />
     </Flex>
   </Padding>
@@ -28,10 +28,10 @@ const SpecialitiesAndServices = ({ specialities, services }) => (
   </Padding>
 )
 
-const RightHandText = ({ title, specialities, services }) => (
+const RightHandText = ({ title, specialities, services, as = 'h2' }) => (
   <Fragment>
     <Padding bottom={0.5}>
-      <SectionTitle>{title}</SectionTitle>
+      <SectionTitle as={as}>{title}</SectionTitle>
     </Padding>
     <FlexCol px={[0, 0, 0, 0, 0, 0, 0]} width={[0, 0, 0, 0, 1, 5 / 6, 5 / 6]}>
       <Padding bottom={1}>
@@ -44,12 +44,12 @@ const RightHandText = ({ title, specialities, services }) => (
   </Fragment>
 )
 
-const CaseStudyHero = ({ caseStudy }) => {
+const CaseStudyHero = ({ caseStudy, as }) => {
   const { posterImage, title, specialities, services } = caseStudy
   return (
     <CaseStudyLayout
       posterImage={posterImage}
-      headings={<SectionTitle>{title}</SectionTitle>}
+      headings={<SectionTitle as={as}>{title}</SectionTitle>}
       textBelowImage={
         <SpecialitiesAndServices
           specialities={specialities}
@@ -59,6 +59,7 @@ const CaseStudyHero = ({ caseStudy }) => {
       rightHandText={
         <RightHandText
           title={title}
+          as={as}
           specialities={specialities}
           services={services}
         />
--- a/src/components/Footer/FooterLinks.js
+++ b/src/components/Footer/FooterLinks.js
@@ -66,7 +66,10 @@ const FooterLinks = () => (
             <Social>
               {social.map(link => (
                 <li key={link.label}>
-                  <ExternalAnchor href={link.to}>
+                  <ExternalAnchor
+                    href={link.to}
+                    title={`yld ${link.label} Account`}
+                  >
                     <img src={link.img} alt={link.label} />
                   </ExternalAnchor>
                 </li>
@@ -76,7 +79,7 @@ const FooterLinks = () => (
           <Col width={[1, 1, 1, 1, 1, 1, 7 / 12]}>
             <GDPR>
               {gdpr.map(link => (
-                <LinkUnderline key={link.label} to={link.to}>
+                <LinkUnderline key={link.label} to={link.to} title={link.label}>
                   {link.label}
                 </LinkUnderline>
               ))}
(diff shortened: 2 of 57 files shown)


src/components/Common/CaseStudyCards/CaseStudyHero.js
```diff
@@ -28,10 +28,10 @@ const SpecialitiesAndServices = ({ specialities, services }) => (
   </Padding>
 )
 
-const RightHandText = ({ title, specialities, services }) => (
+const RightHandText = ({ title, specialities, services, as = 'h2' }) => (
   <Fragment>
     <Padding bottom={0.5}>
-      <SectionTitle>{title}</SectionTitle>
+      <SectionTitle as={as}>{title}</SectionTitle>
```
AniaMakes: why do we need to define as H2 here, if under the hood it's 
```
const SectionTitleSectionTitle = styled.h2`
  ${SectionTitleStyles}`
```
Is it because when we make Styled components it overrides the H2 aspect?

src/components/Footer/FooterLinks.js
```diff
@@ -66,7 +66,10 @@ const FooterLinks = () => (
             <Social>
               {social.map(link => (
                 <li key={link.label}>
-                  <ExternalAnchor href={link.to}>
+                  <ExternalAnchor
+                    href={link.to}
+                    title={`yld ${link.label} Account`}
```
AniaMakes: is there a guidance on the capitalisation of YLD? Is it YLD or yld? Currently the website homepage has YLD, so it may be best to stay consistent with this?

src/components/Footer/Office.js
```diff
@@ -19,7 +19,9 @@ const Office = ({ name, telephone, email, streetAddress }) => (
       <StyledBodyPrimary itemProp="telephone">{telephone}</StyledBodyPrimary>
       {email ? (
         <StyledBodyPrimary>
-          <a href={`mailto:${email}`}>{email}</a>
+          <a href={`mailto:${email}`} title={`Email yld ${name} Office`}>
```
AniaMakes: the only content issue I have quibble with there, is that you'll end up with "Email yld London - HQ Office" which doesn't sound ideal.

src/components/Header/SideNav/index.js
```diff
@@ -47,9 +48,12 @@ const SideNavItem = ({ item, path }) => {
   } else {
     const { label, to, href } = item
     return (
-      <OuterAnchorItem activeClassName="active" to={to} href={href}>
-        {label}
-      </OuterAnchorItem>
+      <OuterAnchorItem
```
AniaMakes: In the old version you have <Component>{label}</Component>
In the new version that {label} in the middle is missing - don't we need it to actually disable the content of the label to the user?

src/components/Header/SideNav/index.js
```diff
@@ -60,12 +64,12 @@ const SideNav = ({ links, isOpen, onClose, path }) => (
       <CloseButton onClick={onClose} />
     </Flex>
     <ul>
-      <OuterAnchorItem activeClassName="active" to="/">
-        Home
-      </OuterAnchorItem>
-      {links.map((link, idx) => (
-        <SideNavItem item={link} key={idx} path={path} />
-      ))}
+      <OuterAnchorItem activeClassName="active" to="/" label="Home" />
```
AniaMakes: same question as above

src/components/Homepage/events/MobileElements.js
```diff
@@ -19,7 +19,7 @@ export const MobileMeetups = ({ events }) => (
         {events.map(conf => (
           <li key={`${conf.id}`}>
             <Subtitle noPaddingBottom>
-              <ExternalAnchor href={conf.linkToEvent}>
+              <ExternalAnchor href={conf.linkToEvent} title={conf.linkToEvent}>
```
AniaMakes: Do we really want the title to be the same as href (rather than conference name, or something)?

src/components/Homepage/services.js
```diff
@@ -107,13 +107,21 @@ const Services = ({ services }) => (
 
               {service.pageReady ? (
                 <Padding bottom={1.5}>
-                  <StyledLink to={`/${service.slug}`}>Learn more</StyledLink>
+                  <StyledLink
+                    to={`/${service.slug}`}
+                    title={`Learn more - ${service.title}`}
```
AniaMakes: would "Learn more about our ${service.title} service" be more descriptive / better?

src/components/Common/CaseStudyCards/CaseStudyHero.js
omonk: yep, you can pass `as` as a prop to a styled component and it will change the rendered HTML tag

src/components/Footer/FooterLinks.js
omonk: every time I go to write YLD/yld/Yld/etc. I think of this slack conversation 

https://yld.slack.com/archives/C04RXJ4UG/p1552906659073800

omonk: I believe `yld` is the proper way

src/components/Footer/Office.js
omonk: yes, thought of that, it still makes sense though no?

src/components/Homepage/events/MobileElements.js
omonk: good point, will change

src/components/Header/SideNav/index.js
omonk: `{label}` was a child of this component which isn't necessary, it can be included as a props as that is clearer as to what its getting at in the `OuterAnchorItem` component definition. 

It's a bit of personal preference and style tbh, I was playing around with this component and ended up putting it back together like this.

src/components/Header/SideNav/index.js
omonk: same answer as above :D

src/components/Homepage/events/MobileElements.js
omonk: good point, will change

src/components/Homepage/services.js
omonk: updated

src/components/Homepage/events/MobileElements.js
omonk: updated

src/components/Footer/Office.js
AniaMakes: True, it does!

PR #273, opened 2019-06-27
DoctorLink Case study

GBouffard:
## Description - [DoctorLink case study](https://trello.com/c/1y2tUZHp/175-case-study-doctorlink)

- [x] Populate Contentful (There are 31 sections!!)
- [x] Link data to page via GraphQL
- [x] Created new re-usable React components that can potentially be re-used in future case studies
- [x] Render CaseStudyHero, CaseStudyPreview and render data from the 31 sections using the components created above.
- [ ] Handle styling for all the breakpoints

## Review template


src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,578 @@
...
+            'https://www.yld.io/static/logo_animated-832020608244057f6a9d73e80994ac4a.gif'
+        }}
+      />
+      <CaseStudyHero caseStudy={caseStudy} />
+
+      <GreyBackground>
+        <Grid>
+          {/* DoctorLink is a healthcare technology company with a mission to simplify the route to health and wellbeing for patients globally... */}
+          {shouldRender(data1) && (
+            <BlockRow topIndex={'6'} bottomIndex={'6'} widthPerc="75%">
```
omonk: Rows in grids are always 100%, if you want to have a column this width please use the columns defined in the design

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,578 @@
...
+// import Statement from '../../components/Common/Statement'
+import BlueBackground from '../../components/Common/BlueBackground'
+import GreyBackground from '../../components/Common/GreyBackground'
+import CaseStudyPreview from '../../components/Common/CaseStudyCards/CaseStudyPreview'
+
+// const PaddingBottomGrid = styled(Grid)`
+//   padding-bottom: ${({ theme }) => theme.space[7]};
+// `
+
+const BlockRow = styled(Row)`
```
omonk: I understand what you're trying to do with this BlockRow component but ultimately it's going to cause you more  issues in the long run.

Please make specific rows for each section, it'll be a lot of components I know but due to the nature of the page, the design is not going to change a huge amount. It's not adding complexity to the page, it's just adding single layer wrappers, of course you end up with more components but it's a lot safer. It's better practise in this situation to create specific components here instead of trying to create a one size fits all solution. 

Due to having this BlockRow component wrapping all of the blocks, one change here has an effect on 32 sections.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,578 @@
...
+  padding-top: ${({ theme, topIndex }) =>
+    topIndex ? theme.space[topIndex] : null};
+  padding-bottom: ${({ theme, bottomIndex }) =>
+    bottomIndex ? theme.space[bottomIndex] : null};
+  flex-direction: ${({ rowReverse }) => `row${rowReverse ? '-reverse' : null}`};
```
omonk: because flex-direction defaults to  `row` you can you just conditionally add reverse

`${({ rowReverse })=> rowReverse ? 'flex-direction: 'row-reverse' :  null}`

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,578 @@
...
+    padding-bottom: ${({ theme }) => theme.space[7]};
+`}
+
+${breakpoint('desktop')`
+    padding-bottom: ${({ theme }) => theme.space[8]};
+`} */
+`
+
+const renderImage = imageUrl =>
+  imageUrl && <img src={imageUrl} alt={imageUrl} />
```
omonk: You'll need to use `<Image/>` from the common components. It's doing everything we already have defined here and also means we can use fluid image types from gatsby that we were discussing yesterday: https://using-gatsby-image.gatsbyjs.org

Please have a look at other places where we're using the gatsby image component, for example the homepage where we're using the `GatsbyContentfulFluid_withWebp` fragment. 

Performance is really important on this site, if we use a normal native HTML `` tag this ends up with the page loading all of the images at once. Using the gatsby-image module, `Img` have lazy loading out of the box.

omonk: Compare the two here in terms of resources size, 134kb compared to  3.4mb 😱 

Gatsby Image
image

Native Img
image

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Cool. Will do then 👍

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: You sure of that? I just tried
`${({ rowReverse }) => (rowReverse ? 'flex-direction: row-reverse' : null)}` and it does not seem to work

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Oh yes. You mentioned it yesterday and I forgot about it. Thanks for solving it 👍 Huge difference. Being lazy is the right way 😆

src/components/Common/CaseStudyCards/CaseStudyBlocks.js
```diff
@@ -0,0 +1,91 @@
...
+  const l = data.length
+  const lWithGap = l + 1
+  const colWidth = [
+    2 / lWithGap,
+    2 / lWithGap,
+    1 / lWithGap,
+    1 / lWithGap,
+    1 / l,
+    1 / l
+  ]
```
omonk: This is the kind of stuff that was I had in mind when advising not to make a one size fits all solution. It certainly works but I can't see the advantage of this over something like 

`{images && images.length && images.map((image) => <Col width={[1, 1, 1, 1, 3 / 12]}><Image {...image}/></Col>)}`

which is where we're using this for

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,689 @@
...
+import CaseStudyPreview from '../../components/Common/CaseStudyCards/CaseStudyPreview'
+
+const BlockRow = styled(Row)`
+  flex-direction: ${({ rowReverse }) => `row${rowReverse ? '-reverse' : null}`};
+  align-items: ${({ alignCenter }) => (alignCenter ? 'center' : null)};
+  padding-top: ${({ theme, mobile }) =>
+    mobile && mobile.top ? theme.space[mobile.top] : null};
+  padding-bottom: ${({ theme, mobile }) =>
+    mobile && mobile.bottom ? theme.space[mobile.bottom] : null};
+  width: ${({ widthPerc }) => widthPerc};
```
omonk: width should not be here

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,689 @@
...
+    ({ genericBlockText, genericBlockImages, ...props }) => ({
+      image: getImage(genericBlockImages, 0),
+      text: genericBlockText && genericBlockText.genericBlockText,
+      ...props
+    })
+  )
+  return normalised[0]
+}
+
+const normaliseImages = ({ genericBlockImages = [] }) => genericBlockImages
```
omonk: This is a bit confusing to understand at first. If you want to write a `get` function with a default I would make it a little clearer. At first I thought it was just returning itself.

```
const getImages = (data) => data.genericBlockImages ? data.genericBlockImages  : []
```

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,689 @@
...
+      <GreyBackground>
+        <Grid>
+          {/* DoctorLink is a healthcare technology company with a mission to simplify the route to health and wellbeing for patients globally... */}
+          {shouldRender(data1) && (
+            <BlockRow
+              mobile={addPaddings('4', '4')}
+              smallTablet={addPaddings('4', '4')}
+              tablet={addPaddings('4', '4')}
+              desktop={addPaddings('6', '6')}
+              widthPerc="75%"
```
omonk: please remove all widths from Row

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,689 @@
...
+          ...caseStudy,
+          socialLogo:
+            'https://www.yld.io/static/logo_animated-832020608244057f6a9d73e80994ac4a.gif'
+        }}
+      />
+      <CaseStudyHero caseStudy={caseStudy} />
+
+      <GreyBackground>
+        <Grid>
+          {/* DoctorLink is a healthcare technology company with a mission to simplify the route to health and wellbeing for patients globally... */}
```
omonk: This section is a `<Statement>` component in Common, please use that!

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,689 @@
...
+          >
+            <TextAndImageBlock data={normalise(data8)} />
+          </BlockRow>
+        )}
+
+        <ReversableDiv>
+          {/* Colour */}
+          {shouldRender(data9) && (
+            <BlockRow
+              mobile={{ bottom: '5' }}
```
omonk: The value story at this key is going to be used in selecting an index from an array, I wouldn't use a string here.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Yep. Will remove. I just didn't take care of it yet.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Good idea. Will do.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: oh yes. Will do.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,689 @@
...
+            <FullWidthBlock data={normalise(data7)} />
+          </BlockRow>
+        )}
+
+        {/* Hi I am Noto sans */}
+        {shouldRender(data8) && (
+          <BlockRow
+            rowReverse
+            alignCenter
+            mobile={addPaddings('5', '5')}
```
omonk: I would just write `{{bottom: 5, top: 5}}`, the `addPaddings` function isn't flexible in that you can't have top and no bottom and is another step that doesn't give us a huge advantage.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Using the `<Statement>` component solved that issue for us. Resolved.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Using the `<Statement>` component solved that issue for us. Resolved.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Done.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Was resolved.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: Yeah. I also thought of that. I'm keen to go with it but we should discuss it too as I want to make sure we won't get into blockers in the future by doing that. I will investigate a bit now. We can discuss it after stand-up.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: I don't understand what you mean here... 🤔

src/components/Common/CaseStudyCards/CaseStudyBlocks.js
GBouffard: Yeah; I will remove the possibility of customising this component. You are right: It might be added logic that we don't need at the moment.

src/pages/case-study/doctorlink-setting-up-a-design-system.js
GBouffard: works with strings!

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,638 @@
...
+    genericBlock31: data31
+  } = caseStudy
+
+  return (
+    <Layout location={location}>
+      <Head
+        page={{
+          ...caseStudy,
+          socialLogo:
+            'https://www.yld.io/static/logo_animated-832020608244057f6a9d73e80994ac4a.gif'
```
omonk: So this can now be `caseStudy.posterImage.file` me thinks

omonk:
🙌 one more minor comment about SEO stuff then im happy!

src/pages/our-work.js
```diff
@@ -43,9 +43,11 @@ const OurWork = ({ data }) => {
   const engineeringCaseStudies = formatCaseStudies(
     allContentfulTemplatedCaseStudy
   )
+
   const designCaseStudies = formatCaseStudies(
     allContentfulNonTemplatedCaseStudy
   )
+
```
omonk: You'll want to add `allContentfulNonTemplatedCaseStudyV2` to this as well

GBouffard: 👍

GBouffard: It's actually already there.

omonk: It is in your GraphQL query yes but the data it returns is not being used in the OurWork component so you're not seeing the doctor link project in the case study list on `/our-work`

```
  const {
    allContentfulTemplatedCaseStudy,
    allContentfulNonTemplatedCaseStudy,
    allContentfulNonTemplatedCaseStudyV2 <--- this bit here is missing in L37
  } = data
```

src/pages/case-study/doctorlink-setting-up-a-design-system.js
```diff
@@ -0,0 +1,674 @@
...
+  VideoBlock,
+  ImagesBlock,
+  TextAndImageBlock,
+  TextAndResizedImageBlock
+} from '../../components/Common/CaseStudyCards/CaseStudyBlocks'
+// import SubtitleWithBody from '../../components/Common/SubtitleWithBody'
+import Statement from '../../components/Common/Statement'
+import CaseStudyPreview from '../../components/Common/CaseStudyCards/CaseStudyPreview'
+
+const BlockRow = styled(Row)`
```
GBouffard: @omonk Would you like me to move the `BlockRow` component to `./src/components/Common/CaseStudyCards/CaseStudyBlocks.js`. Maybe something re-usable for you with Central working?

omonk: No worries for now, I will move a few bits of pieces around in my central working branch after this is merged and I have rebased

src/pages/our-work.js
GBouffard: done

PR #277, opened 2019-07-01
Reuse existing open positions component

SemraSh:
## Description - [Trello ticket](https://trello.com/c/gLUJ8aIO/482-audit-use-of-lever-job-posts-component-make-sure-we-are-re-using-the-same-component-on-homepage-join-us-page)

Make use of existing `OpenPositions` component and removed repetition.

## Review template


SemraSh:
A story is _not_ added for the common component because graphQL queries aren’t being run that are required for this component. 
There are some workarounds with this. 
Please see (this issue created for this)[https://github.com/gatsbyjs/gatsby/issues/12007] .

src/components/Common/OpenPositions.js
```diff
@@ -1,12 +1,10 @@
 import React from 'react'
+import { Padding } from 'styled-components-spacing'
 import { Col, Row } from '../grid'
 import JobsByLocation from '../JobsByLocation'
```
omonk: Just looking at that layout, this `JobsByLocation` is only used in this this file, can you move this to `Common` as well please

src/components/Common/OpenPositions.js
```diff
@@ -1,12 +1,10 @@
...
-import { Padding } from 'styled-components-spacing'
 import { SectionTitle, Subtitle } from '../Typography'
-import { Section } from './elements'
-import Hr from '../Common/Hr'
-import GetInTouch from '../Common/GetInTouch'
-import JobLink from '../Common/JobLink'
+import { Section } from '../JoinUs/elements'
```
omonk: As we're now using this component in the homepage and joinus, we shouldn't rely on components defined in `JoinUs` as it increases obscurity and can cause issues in the future in that changing JoinUS page also affects the homepage.

src/components/Common/OpenPositions.js
```diff
@@ -41,26 +39,20 @@ const renderJobsForlocation = (jobs, location, key) => {
...
-        </Padding>
-      </Section>
-      <GetInTouch title={getInTouchTitle} contactText={getInTouchText} />
-    </div>
-  )
-)
+const OpenPositions = React.forwardRef(({ data: { title } }, ref = null) => (
+  <div ref={ref}>
+    <Section greyBg id="open-positions">
+      <Padding top={{ smallPhone: 3, tablet: 4 }}>
```
omonk: As Anna said in a comment, please remove Padding here.

The reason for removing padding in instances like this is because we're adding HTML to the DOM that isn't necessary. Because we have this wrapper `Section` we can use styled components to add padding without introducing more HTML.

```
const StyledSection = styled(Section)`
   /* all the padding here with breakpoints included */
`
```

This way you go from:
```
<div class="section">
    <div class="padding-element">
        <h2>Title</h2>
    </div>
</div>
```

To:
```
<div class="section">
    <h2>Title</h2>
</div>
```

It might not seem like much but do this 100+ times and you're adding more in general which is not what we want.

As I said "in instances like this", there are instances when Padding might be ok to use. For example say you had something like this: 

```
<div>
    <h2>Title</h2>
    <p>copy section</p>
    <img src="" alt="">
</div>
```

and needed padding on this div and only padding, that might be a decent time to use it but ultimately I don't see the advantage of using `Padding` for a site like ours.

@AniaMakes

src/components/Common/OpenPositions.js
omonk: Looking more closely at this `Section`

```export const Section = ({ children, greyBg, ...props }) => {
  const content = <Grid {...props}>{children}</Grid>

  if (!greyBg) {
    return content
  }

  return <GreyBackground>{content}</GreyBackground>
}```

You can be sure that this section is going to be wrapped in a `GreyBackground` component at all times so lets get rid of the section here and just wrap the entire thing with `GreyBackground`

File: src/components/Common/OpenPositions.js [hunk 4]
```diff
@@ -41,26 +39,20 @@ const renderJobsForlocation = (jobs, location, key) => {
...
-          <Padding top={{ smallPhone: 3, tablet: 4 }}>
-            <Hr />
-          </Padding>
-        </Padding>
-      </Section>
-      <GetInTouch title={getInTouchTitle} contactText={getInTouchText} />
-    </div>
-  )
-)
+const OpenPositions = React.forwardRef(({ data: { title } }, ref = null) => (
```
omonk: We're missing a point from the ticket here

```Create limit prop to display a limited number of jobs on homepage. See @jameshevey3 comment below```

Currently it's set up so that if Nick were to add 20 jobs to lever, they would all display on the homepage which isn't what we want. 

- Add a limit prop to here then pass it to `JobsByLocation`, I would say we can use this variable as a limit filter in the graphql query but unfortunately we need to limit on the location not all lever jobs as a whole.

File: src/components/Homepage/jobs.js [hunk 5]
```diff
@@ -1,68 +1,20 @@
...
-          ))
-        }
-      </JobsByLocation>
-    </Row>
-    <Row>
-      <Col width={[1]}>
-        <Padding top={{ smallPhone: 3, smallTablet: 4 }}>
+        <Padding top={{ smallPhone: 3, smallTablet: 4, desktop: 3.5 }}>
           <StyledLink to="/join-us" title="View all openings">
             View all openings
```
omonk: Following on from [this](https://github.com/yldio/yld.io/pull/277/files#r300232586) comment: 

Something else I've noticed is that even if we only have a maximum of 4 jobs per location, we're still displaying "View all openings" under the jobs list which would be weird for a user to go to view more open positions, only to find that the ones they'd seen already were the only ones we have available.

My initial suggestion was to get the number of jobs and work out if the rendered results in the homepage were capped and display a different message but the process of getting the jobs happens several levels beneath this location sooooo I would opt to just change the messaging here to be something like , "find out more about working with us", I'm not sure, speak to Hana or Greta as my copy writing skills are dire.

src/components/Common/OpenPositions.js
omonk: ```I would say we can use this variable as a limit filter in the graphql query but unfortunately we need to limit on the location not all lever jobs as a whole.```

^ ignore this, I looked a bit closer at the current query in place and it turns out we can do this.

Check it out [here](http://localhost:8000/___graphql?query=query%20JOBS_BY_CITY(%24groupLimit%3A%20Int)%20%7B%0A%20%20allLever(limit%3A%2040)%20%7B%0A%20%20%20%20group(field%3A%20categories___location%2C%20limit%3A%20%24groupLimit)%20%7B%0A%20%20%20%20%20%20edges%20%7B%0A%20%20%20%20%20%20%20%20node%20%7B%0A%20%20%20%20%20%20%20%20%20%20id%0A%20%20%20%20%20%20%20%20%20%20categories%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20location%0A%20%20%20%20%20%20%20%20%20%20%20%20commitment%0A%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%20%20text%0A%20%20%20%20%20%20%20%20%20%20hostedUrl%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A&operationName=JOBS_BY_CITY&variables=%7B%0A%20%20%22groupLimit%22%3A%204%0A%7D)

omonk: ahah im going to back track again, we can't do it this way because we can't pass variables to the StaticQuery from above,  this is a design limitation with Gatsby unfortunately. We will have to transform the data coming back from the query

src/components/Homepage/jobs.js
SemraSh: We agreed with @hanalodhi that we should say `Learn more` instead of `View all openings`. 
Thank you for pointing out.

src/components/Common/OpenPositions.js
```diff
@@ -0,0 +1,64 @@
...
+
+import { Col, Row, Grid } from '../grid'
+import JobsByLocation from './JobsByLocation'
+import { SectionTitle, Subtitle } from '../Typography'
+import JobLink from './JobLink'
+import GreyBackground from './GreyBackground'
+
+export const OpenPositionsWithRef = React.forwardRef((props, ref) => (
+  <div ref={ref}>
+    <OpenPositions {...props} limit={40} />
```
omonk: I would include `limit` in the props here not hard coding it to 40, in its current use we do want *all* of the jobs to be displayed but remember we're in the `Common` directory. In general any reusable components should have a flexible API.

e.g. if we wanted to add a "go to jobs" button on the homepage that would scroll down to a list of roles on the same page, we'd want to use this component but would need to rewrite the props to have a custom limit

src/pages/index.js
```diff
@@ -79,7 +79,6 @@ const IndexPage = ({
       </LatestPosts>
       <GreyBackground>
         <Jobs />
-        <Padding bottom={{ smallPhone: 1.5, smallTablet: 0 }} />
```
omonk: This is removed but we're now missing the spacing in the view. 

`const StyledJobs = styled(Jobs)`{/*...padding...*/}` `

SemraSh: We decided that this Padding is not necessary.

omonk: As Semra pointed out, we're not. My bad

PR #278, opened 2019-07-01
F speciality (Speciality pages should display neatly even with incomplete data)

AniaMakes:
## Description - [Trello ticket 540](https://trello.com/c/OMhOH2ki/540-seo-speciality-template-update)

Aim: Speciality pages should display neatly even with incomplete data

Helper 1: flatten the Speciality GQL data into an object that can be easily compared against a schema
Helper 2: create schemas and use the validation to see if we have the data we need
Helper 3: have fields as compulsory in Contentful (has `// required` in line when creating flattenedSpeciality)

### Required sections:
**Head**
**IntroSection**
**TrainingSection**
**GetInTouch**
>> Ideally the whole Speciality return would be wrapped in a conditional that validates against the schema for flattened Speciality and if that fails, it would instead show a 404. Importing the 404 page  and showing it conditionally doesn't work - maybe we could have design come up with some pretty error we could display if no data can be fetched? Having said that, I made all the fields needed for these sections compulsory on Contentful. 

### Optional sections (don't render if no or dodgy data):
**ProjectsSection** - this section takes `projects` and `clients` arrays and conditionally renders the projects and / or the clients inside (`array && array.length && <Component />`). Before even rendering this component the parent checks if it has a valid data for either projects or clients. If it doesn't, it doesn't render the ProjectsSection at all
**Community** - all the information for this is in the `flattenedSpeciality` so all this does is it runs a schema validation and if that's all good renders the Community component (if not, it doesn't)
**BlogListing** - the initial `posts` array is explicitly defined as either whatever data there is or an empty array. Inside BlogListing it maps over the array (so [ ] remains [ ]) and if there's no listings it will render only link to the medium blog
**Tutorials** - fairly simplistic `array && array.length && <Component />`
**Books** - always attempts to render and the `array && array.length && <Component />` is *inside* the BooksSection component. Given it's only used here, I'm leaving it as it is (unless people think I should move it for consistency)

Automates:
**Events** - this is OK as is - the `specialityEvents` function returns either an array of events or an empty array, the component is displayed either way, if there are no events it will display a helpful message.

## Review template


src/components/Speciality/Community.js
```diff
@@ -26,18 +26,28 @@ const ImgContainerCol = styled(Col)`
...
+        <TrimmedImage
+          alt={communityBackgroundTitle}
+          src={communityBackgroundUrl}
+        />
         <Padding top={5} bottom={5}>
           <Row>
             <ImgContainerCol width={[1, 1, 1, 1, 6 / 12]}>
-              {logo ? (
+              {communityLogoTitle ? (
                 <img
```
GBouffard: I'm surprised we don't use the Common/Image.js component here as it accepts an image object with title and file keys (file also having url as a key).
Using ing works fine if we pass the values of `alt` and `src` directly but I don't see a change in the data being passed. Is this coming from GraphQL somewhere.

src/components/Speciality/Projects.js
```diff
@@ -84,7 +84,7 @@ const ProjectsSection = ({ related, title, clients }) => (
           <Col width={[1, 1, 1, 1, 1 / 2]} />
         </Row>
       ) : null}
-      {clients ? <CompaniesHelped clients={clients} /> : null}
+      {clients && clients.length ? <CompaniesHelped clients={clients} /> : null}
```
GBouffard: While we refactor, in general I think it's always best to write this as:
`{clients && clients.length && <CompaniesHelped clients={clients} />}`
It removes a `: null`that we don't need.

omonk:
image
whilst we're here can you fix this issue,  [ticket](https://trello.com/c/YyEDv0UM/509-why-seo-component-mobile-bug)

src/components/Speciality/Intro.js
```diff
@@ -41,14 +41,14 @@ const IntroSection = ({ speciality }) => (
...
               </BodyPrimary>
             </Flex>
           </Col>
           <Col width={[1, 1, 1, 1, 6 / 12]}>
             <img
-              alt={speciality.introGraphic.title}
-              src={speciality.introGraphic.file.url}
+              alt={speciality.introGraphicTitle}
+              src={speciality.introGraphicFileUrl}
               style={{ maxHeight: '100%' }}
```
omonk: Whilst we're here, lets update this intro section to use RatioContainer with the width and height set to `100`

Also, go back to the query and update it to use `GatsbyContentfulFluid_withWebp` not the `file {  title, url }`. 
```
fluid(maxWidth: 30) {
    ...GatsbyContentfulFluid_withWebp
}
```

src/templates/speciality.js
```diff
@@ -38,6 +43,644 @@ const getSpecialityEvents = (title, events) =>
...
+//       type: 'string',
+//       title: 'The Contacttext Schema',
+//       default: '',
+//       examples: [
+//         'Whether you need help with a project or are interested in training, we love to talk about all things React.'
+//       ],
+//       pattern: '^(.*)$'
+//     }
+//   }
+// }
```
omonk: Can we remove this?

src/components/Speciality/Community.js
omonk: @GBouffard where you're wondering about the formation of the title and url I believe this is coming from the transformation in `speciality.js`

On a whole Guilluame is right, we should be using the `Image` component from `Common/` but we should also be using the `GatsbyContentfulFluid_withWebp` format of the image itself not just a straight up file.

omonk: And also `RatioContainer`

src/components/Speciality/Training.js
```diff
@@ -5,14 +5,14 @@ import { Padding } from 'styled-components-spacing'
 import GreyBackground from '../Common/GreyBackground'
 import StyledLink from '../Common/StyledLink'
 
-const TrainingStage = ({ title, body, icon }) => (
+const TrainingStage = ({ title, body, iconTitle, iconUrl }) => (
   <Col width={[1, 1, 1, 1, 4 / 12]}>
     <Padding top={4} bottom={4}>
       <Padding bottom={1.5}>
-        <img src={`https://${icon.file.url}`} alt={icon.title} />
+        <img src={`https://${iconUrl}`} alt={iconTitle} />
```
omonk: I'm so confused as to why we're writing `https://` in here 🤔 this should be removed. For example, what if the url changes to `https://path.to.image` and we then get `https://https://` 👎

src/components/Speciality/Training.js
```diff
@@ -5,14 +5,14 @@ import { Padding } from 'styled-components-spacing'
 import GreyBackground from '../Common/GreyBackground'
 import StyledLink from '../Common/StyledLink'
 
-const TrainingStage = ({ title, body, icon }) => (
+const TrainingStage = ({ title, body, iconTitle, iconUrl }) => (
   <Col width={[1, 1, 1, 1, 4 / 12]}>
     <Padding top={4} bottom={4}>
       <Padding bottom={1.5}>
```
omonk: Please remove all `Padding` from this file

src/components/Speciality/Training.js
omonk: Again this should also use the image component from Common and the `GatsbyContentfulFluid_withWebp`

src/templates/speciality.js
```diff
@@ -65,22 +708,31 @@ const Speciality = ({
   const books = getExternalType(speciality, `Book`)
```
omonk: As we're already doing a big transformation, you could probably move all of these methods to be included in the `flattenSpeciality` method. That way you can just pass `flattenedSpeciality` to the components

src/templates/speciality.js
```diff
@@ -94,14 +746,19 @@ const Speciality = ({
       />
       {tutorials && tutorials.length ? (
         <TutorialsSection speciality={title} tutorials={tutorials} />
-      ) : null}
+      ) : (
+        undefined
+      )}
```
omonk: Everywhere else in this return we're doing `x && x.length && Component` - let's do the same here

src/templates/speciality.js
```diff
@@ -65,22 +708,31 @@ const Speciality = ({
...
-        text={communityText}
-        title={title}
-      />
+      <Head page={flattenedSpeciality} />
+      <IntroSection speciality={flattenedSpeciality} />
+      {!!(
+        (relatedProjects &&
+          relatedProjects.length &&
+          validateProjects(relatedProjects)) ||
+        (clients && clients.length && validateClients(clients))
```
omonk: Where we validate the schema like this, do we not call `.validate(obj)` on the method returned from `ajv.comile()`? instead of just calling the return function...

omonk: Looks like you can also just do `ajv.validate(schema, data);`

omonk: Also, looks like you don't need `clients && clients.length` as ajv will handle all of that for us

https://runkit.com/omonk/5d1acb94846c9f00132410b7

src/components/Speciality/Projects.js
```diff
@@ -36,12 +36,12 @@ const PosterLinks = ({ project }) => (
...
 )
 
-const CompaniesHelped = ({ clients, noOther }) => (
+const CompaniesHelped = ({ clients }) => (
   <Fragment>
     <Row>
       <Col width={[1, 1, 1, 1, 1 / 2]}>
         <Padding top={5} bottom={3}>
-          <Subtitle>{noOther ? 'C' : 'Other c'}lients we helped</Subtitle>
+          <Subtitle>Other clients we helped</Subtitle>
```
omonk: I removed this as `noOther` is not used at all and this ternary string manipulation is not so great.

omonk:
Generally all looking good. I would say that even though we have sections as `required` in contentful and marked as `required` in the transformation function, this still opens ourselves up to issues if content types change. I don't think it's worth spending too much extra time over this but overall this is really good stuff, a few bits of pieces to tidy up but apart from that we're there. Well done 🙌

src/components/Speciality/Projects.js
AniaMakes: Actually, I have a feeling why that was like that - sometimes (not sure exactly why) if you use the `array && array.length && <Component>` approach, the logic will end up rendering literal 0 on the website in place of the component (I've had this issue).

src/templates/speciality.js
AniaMakes: Actually, I have a feeling why that was like that - sometimes (not sure exactly why) if you use the array && array.length && <Component> approach, the logic will end up rendering literal 0 on the website in place of the component (I've had this issue). I'll change it for consistency.

src/templates/speciality.js
AniaMakes: So... you'd this it'd would work, but if I try your suggested approach (to just rely on ajv validation), I get this if clients and relatedProjects are empty arrays (which I guess is unlikely to happen but it's one way of simulating no data)

image

AniaMakes: Should I assume that there will either be no data at all (ie no empty array), or there will be data? Given those assumptions we can just use the schema validation as conditional.

src/components/Speciality/Training.js
AniaMakes: No more Padding here!

src/components/Speciality/Intro.js
AniaMakes: I've pushed this change, it doesn't look to be breaking anything...

src/components/Speciality/Training.js
AniaMakes: done.

src/templates/speciality.js
AniaMakes: done.

src/components/Speciality/Community.js
AniaMakes: Good point, done!

src/templates/speciality.js
AniaMakes: I was right (see below!). I used `{!!(tutorials && tutorials.length) &&
        <TutorialsSection speciality={title} tutorials={tutorials} />
      }` instead. 

image

src/components/Speciality/Projects.js
AniaMakes: I changed this per your suggestion - the padding ain't pretty when one of them is missing, but I guess that's a design input that's needed as part of the conditional colours.

src/templates/speciality.js
AniaMakes: actually, I'd argue against it because then you'd still need to pass either `flattenedSpeciality.talks` OR you'd need to destructure the whole object and why do that when only one key is needed?

omonk: Indeed but what I'm getting at is that we're mixing where our data is coming from. 

`flattenedSpeciality` should become the main source of all of the data, here is where we're going a lot preparation to pass the data in to the components. So by leaving the `getExternalType` method to extract data from the query, you bypass the transformation and thus reduce the reliability of the data created.

`why do that when only one key is needed` 
It doesn't matter, we should follow our own rules we've decided and maintain consistency.

omonk: `props -> transformation -> transformedProps -> data passed to  components` 

compared to 
`props -> transformation -> transformedProps -> data passed to  components` 
plus
`props -> extractionMethod -> data passed to components`

src/templates/speciality.js
omonk: ahh I see what you mean, yep that makes sense.

if there are no related projects it will be `null`, remember GraphQL will return null for anything like `[]`

[query](http://localhost:8000/___graphql?query=%7B%0A%20%20contentfulSpeciality(slug%3A%20%7Beq%3A%20%22rxjs%22%7D)%20%7B%0A%20%20%20%20slug%0A%20%20%20%20relatedProjects%20%7B%0A%20%20%20%20%20%20...%20on%20ContentfulTemplatedCaseStudy%20%7B%0A%20%20%20%20%20%20%20%20__typename%0A%20%20%20%20%20%20%20%20title%0A%20%20%20%20%20%20%20%20slug%0A%20%20%20%20%20%20%20%20introSentence%0A%20%20%20%20%20%20%20%20posterColor%0A%20%20%20%20%20%20%20%20posterImage%20%7B%0A%20%20%20%20%20%20%20%20%20%20title%0A%20%20%20%20%20%20%20%20%20%20file%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20url%0A%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20...%20on%20ContentfulNonTemplatedCaseStudy%20%7B%0A%20%20%20%20%20%20%20%20__typename%0A%20%20%20%20%20%20%20%20title%0A%20%20%20%20%20%20%20%20slug%0A%20%20%20%20%20%20%20%20posterColor%0A%20%20%20%20%20%20%20%20posterImage%20%7B%0A%20%20%20%20%20%20%20%20%20%20title%0A%20%20%20%20%20%20%20%20%20%20file%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20url%0A%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A&operationName=undefined)

src/templates/speciality.js
omonk: `talks.length > 0` will prevent that

src/templates/speciality.js
omonk: The reason for `flattenSpeciality` is to generate data in a shape that is flatten and assign defaults but most importantly something we can trust right?

For example - what happens if we request `rxjs` speciality but it doesn't have any externalResources returned from the query? [it breaks](http://localhost:8000/___graphql?query=%7B%0A%20%20contentfulSpeciality(slug%3A%20%7Beq%3A%20%22rxjs%22%7D)%20%7B%0A%20%20%20%20slug%0A%20%20%20%20externalResources%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A&operationName=undefined), because we're trying to filter `speciality.externalResources`, which as you can see from the query response is `null`. What would happen if we created `externalResources` on `flattenedSpeciality` like this `Get('externalResources', speciality, [])`, we'd be ok and talks, books and projects wouldn't render successfully.

The main aim of this PR is to handle these types of issues, tying up potential bugs so that contentful can give us whatever it wants and we don't have broken pages or deployment errors.

src/templates/speciality.js
```diff
@@ -47,16 +690,16 @@ const Speciality = ({
...
-    relatedProjects,
-    title,
-    clients,
-    communityBackground,
-    communityLogo,
-    communityText,
-    eventIcon,
-    contactText
-  } = speciality
+  const { relatedProjects, clients, eventIcon } = speciality
```
omonk: Following on from [these comments](https://github.com/yldio/yld.io/pull/278/files/d258e94bde8561ed06734668c5159e2106242fe3#r299785181)
I would argue this should be placed in the `flattenSpeciality` method.

src/components/Speciality/Intro.js
```diff
@@ -17,22 +19,39 @@ const StyledBlueBackground = styled(BlueBackground)`
   margin-top: -${remcalc(36)};
 `
 
+const StyledIntroRectangleInner = styled.div`
+  padding-top: ${({ theme }) => theme.space[3]};
+  padding-bottom: ${({ theme }) => theme.space[3]};
+
+  ${breakpoint('smallPhone', 'smallTablet')`
```
omonk: you shouldn't need to  specify smallTablet here as the media queries go up

AniaMakes: *With 'smallTablet'*
image

*Without*
image

src/templates/speciality.js
AniaMakes: Ok, I pushed a thing which I think matches your expectations.

src/templates/speciality.js
AniaMakes: OK, done.

src/components/Speciality/Intro.js
GBouffard: Mmh. That's strange. I guess some other style applies on smallTablet then 🤔

src/components/Speciality/Projects.js
```diff
@@ -52,7 +52,7 @@ const CompaniesHelped = ({ clients, noOther }) => (
 const ProjectsSection = ({ related, title, clients }) => (
   <Grid>
     <Padding top={5} bottom={5}>
-      {related ? (
+      {!!(related && related.length) && (
         <Row>
```
GBouffard: You don't need the `!!` here.

src/components/Speciality/Projects.js
```diff
@@ -83,8 +83,8 @@ const ProjectsSection = ({ related, title, clients }) => (
           )}
           <Col width={[1, 1, 1, 1, 1 / 2]} />
         </Row>
-      ) : null}
-      {clients ? <CompaniesHelped clients={clients} /> : null}
+      )}
+      {!!(clients && clients.length) && <CompaniesHelped clients={clients} />}
```
GBouffard: same

src/components/Speciality/Projects.js
AniaMakes: It's preventative to avoid this: https://github.com/yldio/yld.io/pull/278#discussion_r299540859 !

PR #284, opened 2019-07-05
Doctorlink case study

GBouffard:
## Description - [DoctorLink case study](https://trello.com/c/1y2tUZHp/175-case-study-doctorlink)

- [x] Populate Contentful (There are 31 sections!!)
- [x] Link data to page via GraphQL
- [x] Created new re-usable React components that can potentially be re-used in future case studies
- [x] Render CaseStudyHero, CaseStudyPreview and render data from the 31 sections using the components created above.
- [x] Handle styling for all the breakpoints

## Review template


src/components/Common/CaseStudyCards/CaseStudyBlocks.js
```diff
@@ -19,14 +19,23 @@ const StyledColumnImage = styled(Col)`
 const renderImage = data => data && <Image image={data} />
 
 /* The reason for this is to solve inconsistent CSS Styling Between gatsby develop and build.
-This solution ensures that strong is always bold for any environment */
+This solution ensures that strong is always bold and white for any environment */
 const StyledBodyPrimary = styled(BodyPrimary)`
+  color: ${({ theme, bpColorReverse }) =>
+    theme.colors[bpColorReverse ? 'opacityWhite' : 'white']};
+
```
GBouffard: Using `opacity: ` didn't work here as it applied it to a whole paragraph, including `strong`. Adding an opacity in strong would applied double opacities. I had to use `color` to solve. I added the equivalent of `white opacity 0.5` in theme.

src/components/Common/CaseStudyCards/CaseStudyBlocks.js
```diff
@@ -0,0 +1,172 @@
...
+const TextAndImageBlock = ({
+  data: { text, image },
+  colWidthOne = [1, 1, 1, 1, 1 / 2],
+  colWidthTwo = [1, 1, 1, 1, 1 / 2],
+  middleColWidth = null,
+  colorReverse = false
+}) => (
+  <Fragment>
+    <Col width={colWidthOne}>{renderText(text, colorReverse)}</Col>
+    {text && image && <StyledBreakpointMobilePadding width={[1]} />}
```
omonk: Can you not just  pass `width={[0, 0, 0, 1]}` to here? It'll go to the the `Col` as props

src/components/Footer/index.js
```diff
@@ -30,15 +30,19 @@ const showGetInTouch = pathname => {
...
   const path = getPathnameWithSlashes(pathname)
-  /**
-   * These pages all have a final white component so we make sure
-   * to add a grey background to give them some distinction
-   */
-  const colors = {
-    greyBg: ['engineering', 'design', 'training ', 'about-us', 'open-source']
-  }
 
-  return Object.keys(colors).find(key => colors[key].includes(path))
```
omonk: Why has this been changed so drastically? It's turned it into a binary 'greyBg' or 'white' result, whereas previously it should be extended to include any colour...

If you simply added 'case-study' to the `greyBg` value in the `colors` object you'd have achieved the same thing.

GBouffard: It actually didn't work. I tried to pass 'case-study' in the array.

src/components/Common/CaseStudyCards/CaseStudyBlocks.js
GBouffard: Oh yeah! That works too! didn't think of that!

src/components/Footer/index.js
GBouffard: Are we sure that this was working in the first place? I was surprised that passing 'case-study' in the array didn't work.

omonk: ok my bad, but because the slug isn't just `case-study`. You can switch `colors[key].includes(path)` to `path.includes(colour[key])` and it  should work

PR #289, opened 2019-07-11
Career framework

omonk:
## Description - [Trello ticket](https://trello.com/c/wXeoSqO8/260-career-framework-groomed)

Creates career framework page and corresponding components

### DIFF
--- a/src/components/CareerFramework/Discipline.js
+++ b/src/components/CareerFramework/Discipline.js
@@ -1,128 +1,13 @@
 import React, { Fragment } from 'react'
 import generate from 'shortid'
-import StyledLink from '../Common/StyledLink'
 import styled from 'styled-components'
-import Get from 'lodash.get'
 
-import { Grid, Row, Col } from '../grid'
-import { BodyPrimary, SectionTitle } from '../Typography'
-import Hr from '../Common/Hr'
+import { Grid } from '../grid'
 import GreyBackground from '../Common/GreyBackground'
 import BlueBackground from '../Common/BlueBackground'
-import Image from '../Common/Image'
 
-// levels {
-//     title
-//     info
-//     levelInfo {
-//       id
-//       ctaTitle
-//       ctaUrl
-//       info {
-//         id
-//         content {
-//           content {
-//             value
-//           }
-//         }
-//       }
-//     }
-//   }
-
-const LevelInfo = ({ ctaTitle, ctaUrl, info, title }) => {
-  const infoContent = Get(info, 'content[0].content[0].value', undefined)
-  return (
-    <LevelInfoWrapper>
-      <BodyPrimary noPadding strong>
-        {title}
-      </BodyPrimary>
-      {infoContent && <BodyPrimary>{infoContent}</BodyPrimary>}
-      <StyledLink to={ctaUrl} title={`Learn More about ${title}`}>
-        {ctaTitle}
-      </StyledLink>
-    </LevelInfoWrapper>
-  )
-}
-
-const LevelOrder = styled.div`
-  border: 3px solid ${({ theme }) => theme.colors.black};
-  display: inline-block;
-  font-weight: bold;
-  width: 36px;
-  height: 36px;
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  border-radius: 50%;
-`
-
-const LevelInfoWrapper = styled(Col)`
-  padding-bottom: ${({ theme }) => theme.space[4]};
-`
-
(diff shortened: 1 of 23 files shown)


src/components/CareerFramework/Discipline.js
```diff
@@ -0,0 +1,37 @@
...
+
+import { Grid } from '../grid'
+import GreyBackground from '../Common/GreyBackground'
+import BlueBackground from '../Common/BlueBackground'
+
+import Group from './Group'
+import Join from './Join'
+
+const Wrapper = styled.div`
+  ${({ isActive }) => (isActive ? 'display: block' : 'display: none')}
```
GBouffard: same thing but I'd write as `display: `${isActive ? 'block' : 'none'}`

src/components/CareerFramework/Group.js
```diff
@@ -0,0 +1,23 @@
+import React, { Fragment } from 'react'
+import generate from 'shortid'
+
+import Hr from '../Common/Hr'
+
+import Level from './Level'
+
+const Group = props => {
+  const { levels } = props
```
GBouffard: Can make it one line as `const Group = ({ levels }) => {`

GBouffard: and even the need for the keyword `return`
`const Group = ({ levels }) => levels && levels.length && levels.map(( \\ rest of the code`

src/components/CareerFramework/Join.js
```diff
@@ -0,0 +1,38 @@
...
+
+import Image from '../Common/Image'
+
+const JoinContentCol = styled(Col)`
+  padding-top: ${({ theme }) => theme.space[7]};
+  padding-bottom: ${({ theme }) => theme.space[7]};
+`
+
+const Join = props => {
+  const { title, ctaTitle, ctaUrl, content, image } = props
```
GBouffard: Same comment as above
`const Join = ({ title, ctaTitle, ctaUrl, content, image }) =>  <Grid> \\ rest of the code`

src/components/Common/Head.js
```diff
@@ -22,7 +22,7 @@ const Head = ({ page }) => {
         const { siteTitle, image, siteUrl } = siteMetadata
 
         const title = `${siteTitle} ${
-          page.title === 'YLD' ? '' : ` - ${page.title}`
+          page.title === 'YLD' || !page.title ? '' : ` - ${page.title}`
         } ${
```
GBouffard: same logic but nicer I think:
```
page.title && page.title === 'YLD' ? ` - ${page.title}` :  ''
```

src/components/CareerFramework/Discipline.js
```diff
@@ -0,0 +1,49 @@
...
+import Group from './Group'
+import Join from './Join'
+
+const StyledGrid = styled(Grid)`
+  padding-bottom: ${({ theme, isLastGroupWithoutJoins }) =>
+    isLastGroupWithoutJoins ? theme.space[7] : null};
+`
+
+const Discipline = ({ groups, joins = [] }) => (
+  <>
```
GBouffard: I have a question: What is the advantage of using something like `<>`as opposed to `<Fragment>` here?

src/components/Common/StyledLink.js
```diff
@@ -26,17 +26,31 @@ export const StyledLinkCss = css`
     margin-top: ${remcalc(6)};
     background: ${({ theme }) => theme.colors.text};
     box-sizing: border-box;
+    transition: all ${props => props.theme.animations.fast} ease-out;
```
GBouffard: `${({ theme }) => theme.animations.fast} ease-out;` to keep it consistent with the rest of the file

src/components/Common/StyledLink.js
```diff
@@ -55,6 +69,10 @@ export const StyledLinkCss = css`
     }
   }
 
+  ${is('muted')`
+    color: ${props => props.theme.colors.textLight};
```
GBouffard: same here.

src/components/CareerFramework/Level.js
```diff
@@ -0,0 +1,77 @@
...
+  </LevelInfoWrapper>
+)
+
+const LevelOrder = styled.div`
+  border: 3px solid ${({ theme }) => theme.colors.black};
+  display: inline-block;
+  font-weight: bold;
+  width: 36px;
+  height: 36px;
+  display: flex;
```
GBouffard: A convention is to not define `px` and use 
```import remcalc from 'remcalc'
width: ${remcalc(36)}; instead
`

PR #292, opened 2019-07-15
Featured doctor link case study

GBouffard:
## Description - [Make doctorlink case study the featured design case study](https://trello.com/c/8afMpwd9/564-make-doctorlink-case-study-the-featured-design-case-study-on-the-design-landing-page-related-case-study-on-other-design-case-stu)

- [x] Add missing `h1` on doctor link
- [x] Allow services.caseStudies to accept a case study of type ContentfulNonTemplatedCaseStudyV2.
I added DoctorLink in that array in Contentul (which only has Canon so far). By updating to index 1 it renders the DoctorLink CS. If the intention is to always render 1 case study on Top, I suggest to leave index 0 and have a limitiation of `no more than 1` in Contentful, like it's done for related case studies.
I suggest that PR not to be merged before we have a chance to discuss this with the team.

nb:
Speaking of `services.relatedCaseStudy`, I tried and am still trying to add ContentfulNonTemplatedCaseStudyV2 as a fragment within the query. It's not having it. In Contentful related case study is set-up to potentially be of type ContentfulNonTemplatedCaseStudyV2 so I'm not sure what the problem is. I cannot update in Contentful as it would also update in production unlike services.caseStudies where I can control what to display with the index. It's out of the scope of the ticket but it makes sense to tackle related case Studies as v2 so we can also use them.

## Review template

### DIFF
--- a/src/pages/case-study/doctorlink-setting-up-a-design-system.js
+++ b/src/pages/case-study/doctorlink-setting-up-a-design-system.js
@@ -199,7 +199,7 @@ const IndexPage = props => {
           socialLogo: (caseStudy.posterImage.file || {}).url
         }}
       />
-      <CaseStudyHero caseStudy={caseStudy} />
+      <CaseStudyHero caseStudy={caseStudy} as="h1" />
       <GreyBackground>
         {/* DoctorLink is a healthcare technology company */}
         {shouldRender(data1) && <Statement>{normalise(data1).text}</Statement>}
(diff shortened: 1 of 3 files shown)


omonk:
Remember that this is a template so changes on one page will affect other pages throughout the site. 

e.g. engineering and training now have no case studies because `[0]` in the case studies array is undefined.

image

PR #293, opened 2019-07-16
Central working v2

omonk:
## Description - [Trello ticket](https://trello.com/c/2W01i4Ib/539-case-study-central-working-groomed)

- Add new case study for central working
- Move several methods and components from DoctorLink case study into common components


src/pages/case-study/central-working-changing-how-you-work.js
```diff
@@ -0,0 +1,385 @@
...
+
+const Block3Col = styled(Col)`
+  padding-top: ${({ theme }) => theme.space[3]};
+`
+
+const Block4ImageCol = styled(Col)`
+  padding-top: ${({ theme }) => theme.space[4]};
+`
+
+const Block6ImageCol = styled(Col)`
```
GBouffard: Block4ImageCol & Block6ImageCol are the same. I would use a single component and think of another name

src/pages/case-study/central-working-changing-how-you-work.js
```diff
@@ -0,0 +1,385 @@
...
+  padding-top: ${({ theme }) => theme.space[4]};
+`
+
+const Block7ImageCol = styled(Col)`
```
GBouffard: You can use the new name as above and use `width={[]}` prop with a custom array to render or not depending on breakpoints

src/pages/case-study/central-working-changing-how-you-work.js
```diff
@@ -0,0 +1,385 @@
...
+  ${breakpoint('smallPhone', 'phone')`
+    padding-top: ${({ theme }) => theme.space[4]};
+  `}
+`
+
+const Block9ImageCol = styled(Col)`
```
GBouffard: Same comment as above. You can show or not with a custom `width={[]}`

src/pages/case-study/central-working-changing-how-you-work.js
```diff
@@ -0,0 +1,385 @@
...
+  ${breakpoint('smallPhone')`
+      padding-top: ${({ theme }) => theme.space[4]}
+  `}
+`
+
+const StyledHr = styled(Hr)`
+  margin-left: auto;
+  margin-right: auto;
+
```
GBouffard: Does this work with `margin: 0 auto`; ? Nicer if it does. Dont know if some margins top bottom are set on Hr

src/components/Common/TanBackground.js
```diff
@@ -0,0 +1,5 @@
+import styled from 'styled-components'
+
+export default styled.section`
+  background: ${props => props.theme.colors.tanBg};
```
GBouffard: I would use `${({ theme }) => theme.colors.tanBg` for consistency with the rest of the code. It's not a blocker to merge. I like consistency :)

src/pages/case-study/central-working-changing-how-you-work.js
omonk: Hr has margin top and padding bottom, this is why I've set it here explicitly

src/pages/case-study/central-working-changing-how-you-work.js
omonk: I really don't think its necessary to create a shared component here. a `DRY` approach is not applicable in this situation

src/pages/case-study/central-working-changing-how-you-work.js
omonk: This is an image column, I want it to display at all viewpoints but just give it some padding top on mobile down viewsa

src/pages/case-study/central-working-changing-how-you-work.js
omonk: same reply as comment above 😉

src/components/Common/TanBackground.js
omonk: updated and the bluebackground component, nice one!

src/components/Common/CaseStudyCards/CaseStudyBlocks.js
```diff
@@ -183,7 +183,61 @@ const TextAndResizedImageBlock = ({ data: { text, image } }) => (
...
+  padding-top: ${({ theme, desktop }) =>
+    desktop && desktop.top ? theme.space[desktop.top] : null};
+  padding-bottom: ${({ theme, desktop }) =>
+    desktop && desktop.bottom ? theme.space[desktop.bottom] : null};
+`}
+`
+
+const getImage = (blockImages, index) => blockImages && blockImages[index]
+
+const normaliseAll = (arr = []) =>
```
SemraSh: I think we can should use more specific names than `arr` , `genericContents` maybe?

omonk: Using `arr` here just refers to the argument name, what we get out is a an array of generic blocks

src/pages/case-study/central-working-changing-how-you-work.js
GBouffard: ok. Why? Just trying to understand.

src/pages/case-study/central-working-changing-how-you-work.js
GBouffard: ok. If that's the reason we can leave that 👍

src/pages/case-study/central-working-changing-how-you-work.js
GBouffard: Alright. Makes sense then. I was too lazy to check 😆

src/components/Common/CaseStudyCards/CaseStudyBlocks.js
SemraSh: Yeah I understand it represents an array of something and it is an argument, I just think we should be more specific about naming. Up to you.

SemraSh:
LGTM 👍

PR #294, opened 2019-07-16
Speciality Page Story with toggle

omonk:
## Description - [Trello ticket](https://trello.com/c/5Rd8fBwJ/542-speciality-update-create-story-for-speciality-component)

- Adds story for the entire speciality page
- Adds toggle functionality so we can hide and show specific sections of the template


src/components/Speciality/Talks.js
```diff
@@ -14,7 +14,7 @@ const TalksSection = ({ talks: allTalks }) => {
   const featured = allTalks.find(({ featured }) => featured)
   const cta = allTalks.find(({ cta }) => cta)
 
-  return talks && talks.length ? (
+  return (
```
GBouffard: Why did you remove `talks && talks.length`? Could have been used as `talks && talks.length && <BlueBackground>` with `talks = []` in props

src/templates/speciality-component.js
```diff
@@ -0,0 +1,118 @@
...
+  const tutorials = getExternalType(flattenedSpeciality, `Tutorial`)
+  const books = getExternalType(flattenedSpeciality, `Book`)
+  const specialityEvents =
+    events && events.length > 0 ? getSpecialityEvents(title, events) : []
+
+  const renderProjectsSection =
+    validateProjects(relatedProjects) || validateClients(clients)
+  const renderCommunitySection = validateCommunity(flattenedSpeciality)
+
+  const renderTalksSection = talks && talks.length > 0
```
GBouffard: I would have called all these sections `hasTalks`, `hasBooks` and `hasBlogs` even if it's pretty easy to understand what it does.

src/components/Speciality/Books.js
```diff
@@ -48,66 +48,65 @@ const BooksBox = styled(ExternalAnchor)`
...
-                  </Padding>
-                </BooksBox>
-              </Padding>
-            </BookCol>
-          ))}
-        </Row>
-        <Row>
-          <Col width={[1]}>
+const BooksSection = ({ books, title }) => (
+  <Grid>
```
GBouffard: Same comment as below. Always a good safety net to have `books && books.length &&` with `books = []` in props

src/components/Speciality/Talks.js
omonk: Technically due to the updates in the `speciality-component` we don't need these but you're right they should still be localised here if we're not setting any defaults as well

PR #295, opened 2019-07-16
Learnerbly case study

GBouffard:
## Description - [Learnerbly case study](https://trello.com/c/PRDYKvCC/184-case-study-learnerbly)

(branched from Central Working PR)
- render case study

PR pointing to `central-working-v2` for now to see the differnece
@ollie as you will see I added new backgrounds as well; just like you had to for CW;.
Just so you know where I got the name "regalBlue" and "Mountain Meadow" as it's close to the colors we have in Zeppelin, I used that:
https://www.htmlcsscolor.com/hex/143B52
https://www.htmlcsscolor.com/hex/14B48F

I mean I don't know how else I would have come up with "Mountain Meadow". Didn't even cross my mind!
We can obviously change the name if not happy with it.

All paddings and small bits need to still be done. So far rendering all the blocks in a rough away.

## Review template


src/components/Common/Statement.js
```diff
@@ -26,13 +26,23 @@ const Link = styled(StyledLink)`
...
+
 const Statement = ({ richText, children, as = 'h2' }) => (
   <GreyBackground>
     <PaddedGrid>
       <Row>
         <Col width={[1, 1, 1, 10 / 12, 10 / 12, 9 / 12]}>
           <DisplayTitle as={as} textLight>
-            {children}
+            {children && mapWithBold(children)}
             {richText &&
```
GBouffard: regarding this twisted logic in `Statement` component; It's a tricky one and trickier solution.

it turns out that encapsulating a text with `__` in Contentful gives a bold text in the preview; but not in our apps.
The logic assumes that a bold text will always have an impair index since a split occurs when we meet __;

In other words, it would not work if someone would right `__bold_text1__ __bold_text2__` in Contentful !
Hopefully no one would because to make everything bold we would have `__bold_text1 bold_text2__` that works.
`__bold_text1__, __bold_text2__` would also work with as `,` would not render as bold.


This is not an idea solution but it looks like the Learnerbly case study is the only place where we do such a thing at the moment.

omonk: I would probably opt for something like [this](https://gist.github.com/omonk/a9b7e5f7d285183027d869f6b04affc1)

We have an issue with the Statement component here as we pass it two different types of data, richtext formatted data from contentful and markdown from `longText`, (I'm not sure why we have two different types as we only use the richtext to render anchors  as `<Link/>` which we can do with react-markdown but aaaaanyway).

It's a bit of big change to a Common component but from doing some checks I believe it might be the preferred solution. Reasons for this are: 
- We're utilising libraries and tools we already have in the project
- If we didn't have these tools already `mapWithBold` is a good solution to this issue but what you've started on is entering the world of syntax parsing, something thats notoriously time consuming and tricky to handle so I would let the react-markdown package handle all of that for us.

Downsides of this solution: 
- You have to pass headers and paragraphs as the same renderer incase different markdown types were passed to a field in contentful.

src/pages/case-study/learnerbly-accelerating-professional-growth.js
```diff
@@ -27,14 +26,6 @@ const Block1Col = styled(Col)`
   padding-bottom: ${({ theme }) => theme.space[2]};
 `
 
-const StyledBodyPrimary = styled(BodyPrimary)`
-  display: none;
-
-  ${breakpoint('tablet')`
-    display: inherit;
-  `}
-`
```
omonk: We can go  about this without having to have duplicate content on the page! 😄

src/pages/case-study/learnerbly-accelerating-professional-growth.js
```diff
@@ -90,12 +81,15 @@ const IndexPage = props => {
...
+          <BlockRow
+            mobile={{ bottom: '4', top: '4' }}
+            tablet={{ top: '6' }}
+            style={{ justifyContent: 'space-between' }}
+          >
             <TextColumnsBlock
               data={normalise(data3)}
               colWidthOne={[1, 1, 1, 1, 5 / 12]}
               colWidthTwo={[1, 1, 1, 1, 6 / 12]}
-              middleColWidth={[0, 0, 0, 0, 1 / 12]}
```
omonk: I missed this during the doctor link pr but if we want to space between columns that don't add up to 12 we can just use `space-between` to achieve this. Remember in general we want to keep the DOM as small as possible.

src/pages/case-study/learnerbly-accelerating-professional-growth.js
```diff
@@ -115,12 +109,15 @@ const IndexPage = props => {
       {/* "How did we go about solving it?" */}
       {shouldRender(data5) && (
         <Grid flex>
-          <BlockRow mobile={{ bottom: '5', top: '4' }} tablet={{ bottom: '6' }}>
+          <BlockRow
+            mobile={{ bottom: '5', top: '4' }}
+            tablet={{ bottom: '6' }}
+            style={{ justifyContent: 'space-between' }}
+          >
```
omonk: same here as above

src/pages/case-study/learnerbly-accelerating-professional-growth.js
```diff
@@ -169,31 +166,34 @@ const IndexPage = props => {
           <BlockRow
             mobile={{ bottom: '5', top: '4' }}
             tablet={{ bottom: '6', top: '6' }}
+            style={{ justifyContent: 'flex-end' }}
```
omonk: this forces the image to go to  the right on smaller screens

src/pages/case-study/learnerbly-accelerating-professional-growth.js
```diff
@@ -169,31 +166,34 @@ const IndexPage = props => {
...
-                  // eslint-disable-next-line
-                  heading: props => <SectionTitle {...props} />,
-                  // eslint-disable-next-line
-                  paragraph: props => <StyledBodyPrimary {...props} />
-                }}
-                source={normalise(data7).text}
-              />
+            <Col width={[1, 1, 1, 1, 1, 4 / 12]}>
+              <Row>
+                <Col width={[1, 1 / 2, 1 / 2, 1 / 2, 1 / 2, 1]}>
```
omonk: Ok so wtf is going on here?

Previously we were hiding BodyPrimary on large then showing on smaller screens. Whilst not a bad solution we do have the tools available that means we don't have to have duplicate content. 

So we've pretty much made something that looks like this:
image

In the image, `[ ]`  represents a `<Col />`. So you can see we've nested like this:
```
<Col>
    <Row>
        <Col>
           Header 
        </Col>
        <Col>
            BodyPrimary
        </Col>
    </Row>
</Col>
<Col>
    <Image />
</Col>
```

src/pages/case-study/learnerbly-accelerating-professional-growth.js
```diff
@@ -243,31 +243,34 @@ const IndexPage = props => {
...
-                disallowedTypes={['heading']}
-                renderers={{
-                  // eslint-disable-next-line
-                  paragraph: props => <BodyPrimary {...props} />
-                }}
-                source={normalise(data9).text}
-              />
+            <Col width={[1, 1, 1, 1, 1, 4 / 12]}>
+              <Row>
+                <Col width={[1, 1 / 2, 1 / 2, 1 / 2, 1 / 2, 1]}>
```
omonk: Same deal here with the comment above

omonk:
image
Missing 36px padding bottom on the body text

omonk:
image
Missing 12px padding-top to paragraph here

omonk:
image
Wrong column widths here on largeTablet

omonk:
image
12px padding-top to the body content here

src/pages/case-study/learnerbly-accelerating-professional-growth.js
GBouffard: oh yes! Indeed! I didn't think about that either. Great call 👍

src/pages/case-study/learnerbly-accelerating-professional-growth.js
GBouffard: Nesting is wicked for this! I didn't think of it but great stuff 👍 I wished I did think of it.

src/components/Footer/index.js
```diff
@@ -46,16 +47,26 @@ const getBackgroundColor = ({ pathname, contactUsBg, is404 = false }) => {
...
-  return Object.keys(colors).find(key => colors[key].includes(path))
+  const path = getPathnameWithSlashes(pathname)
+
+  const isGreySubUrl = Object.keys(GreyBgSubUrls).find(key =>
+    GreyBgSubUrls[key].some(subUrl => path.includes(subUrl))
+  )
+
+  if (isGreySubUrl) {
+    return 'greyBg'
+  }
```
omonk:
```
 const colors = {
    greyBg: [
      'engineering',
      'design',
      'training ',
      'about-us',
      'open-source',
      'case-study'
    ]
  }

return Object.keys(colors).find(key =>
    colors[key].some(subUrl => path.includes(subUrl))
  )
```

This return value should be enough as it will return the key of the colors object in which it finds the path within the subUrl. 

At the moment if colors is something like  `const colors = { white: 'case-study' }` and the path is `case-study/name-of-study` then the conditional `if (isGreySubUrl) { return 'greyBg' }` will be true and make the background colour of the contact us will be `greyBg`

GBouffard: oh yeah. I thought we already did that. Gonna update.

PR #298, opened 2019-07-18
Fix-react-warning-proptypes-title-required

GBouffard:
## Description
We had all these warnings related to propTypes title in multiple pages.
Links seem to work without it so I'm not sure why it didn't show in the first place

- Remove title as required propType and remove propType as a whole for StyledLink component
- Refactor StyledLink.js as a whole.
- Remove title as required propType and remove propType as a whole for ExternalAnchor component

## Review template

### DIFF
--- a/src/components/Common/ExternalAnchor.js
+++ b/src/components/Common/ExternalAnchor.js
@@ -1,5 +1,4 @@
 import React from 'react'
-import PropTypes from 'prop-types'
 
 const ExternalAnchor = ({ children, openInNewWindow = true, ...props }) => {
   return (
@@ -11,10 +10,6 @@ const ExternalAnchor = ({ children, openInNewWindow = true, ...props }) => {
       {children}
     </a>
   )
-}
-
-ExternalAnchor.propTypes = {
-  title: PropTypes.string.isRequired
 }
 
 export default ExternalAnchor
--- a/src/components/Common/StyledLink.js
+++ b/src/components/Common/StyledLink.js
@@ -96,10 +96,16 @@ const StyledLink = ({ external, to, href, children, ...props }) => {
       }
     : {}
 
-  const redirectProps = to ? { to } : { as: 'a', href }
+  if (to) {
+    return (
+      <Anchor {...externalProps} to={to} {...props}>
+        {children}
+      </Anchor>
+    )
+  }
 
   return (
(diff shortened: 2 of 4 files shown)


src/components/Common/ExternalAnchor.js
```diff
@@ -13,8 +12,4 @@ const ExternalAnchor = ({ children, openInNewWindow = true, ...props }) => {
   )
 }
 
-ExternalAnchor.propTypes = {
-  title: PropTypes.string.isRequired
```
omonk: There are a few more areas in the codebase you've missed that I added. Search the repo for 
`title: PropTypes.string.isRequired` and you'll find em

src/components/Common/StyledLink.js
```diff
@@ -96,23 +95,14 @@ const StyledLink = ({ external, to, href, children, ...props }) => {
...
-        {children}
-      </Anchor>
-    )
-  }
+
+  const redirectProps = to ? { to } : { as: 'a', href }
 
   return (
-    <Anchor {...externalProps} as="a" href={href} {...props}>
+    <Anchor {...externalProps} {...props} {...redirectProps}>
```
omonk: I'm umming and erring about this change. It's perfectly fine at the moment and isn't broken so the first thing that comes to mind is "if it aint broke don't fix it". Generally ternaries are considered to take a bit longer to understand than the humble `if () {}` conditional. 

It's up to you with this but personally I'm very happy with what we have at the moment as its clear and works.

GBouffard: `ternaries are considered to take a bit longer to understand than the humble if () {} conditional.`
Mhh. I didn't know that. Shame because I love ternaries ❤️ Let's discuss at stand-up. We may revert them

GBouffard: Just did some reading on the topic and some other devs say that ternary is faster. Ultimately it does not seem to be everyone's opinion but what also comes out is that the speed difference is negligible. I like Ternary readability. Once again, we can discuss at SU

src/components/Common/ExternalAnchor.js
GBouffard: I think I got them all. I left `CaseStudyPreview.js` propTypes as there are loads of them. We may want to check this validation.

PR #303, opened 2019-07-22
Add /blog/ page

SemraSh:
## Description - [Create "Blog" Page](https://trello.com/c/yrC1U1hY/565-create-blog-page)

Creating a brand new blog page with Medium post reviews.
> Currently the `blog` link in the navigation takes users straight to medium, we would like to keep them on the yld.io website for as long as possible.

## Review template


src/pages/blog.js
```diff
@@ -0,0 +1,138 @@
...
+        page={{
+          title: blogPageMeta.title,
+          seoTitle: blogPageMeta.seoTitle,
+          seoMetaDescription: blogPageMeta.description
+        }}
+      />
+      <Grid>
+        <Row>
+          <Col>
+            <PageDescription>
```
omonk: As PageDescription is only using padding top and bottom you can create `<PageDescriptionCol>` and combine the two like 

`const PageDescriptionCol = styled(Col)` ...padding values...`

src/pages/blog.js
```diff
@@ -0,0 +1,138 @@
...
+              <SectionTitle as="h1">{blogPageMeta.title}</SectionTitle>
+              <FixedWidthDisplayTitle regular textLight>
```
omonk: We can use the grid to set widths, from looking at invision I can see its meant to be `9 / 12` on desktop.

src/pages/blog.js
```diff
@@ -0,0 +1,138 @@
...
+                {blogPageMeta.description}
+              </FixedWidthDisplayTitle>
+            </PageDescription>
+          </Col>
+        </Row>
+      </Grid>
+      <Grid>
```
omonk: You don't need to have two Grids here I don't think? Two `Row`s should do it

src/pages/blog.js
```diff
@@ -0,0 +1,138 @@
...
+        <StyledDisplayTitleWrapper>
+          <DisplayTitle>Recent articles</DisplayTitle>
+        </StyledDisplayTitleWrapper>
+        {mediumPosts &&
+          mediumPosts.length > 0 &&
+          mediumPosts.map((mediumPostData, idx) => {
+            const isLastPost = idx === mediumPosts.length - 1
+            return (
+              <Fragment key={idx}>
```
omonk: Using `index` as a key is not advised in React, I need to add this as an eslint rule but please use shortid as we do in other areas. 

More info: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

SemraSh: Yeah I agree, I have seen `index` used as a key in many places in the codebase and I was intending to ask that so left it as is for the time being. I will change that.

src/pages/Untitled-1
```diff
@@ -0,0 +1,7 @@
+- imaji getirmeye calis DONE
+- linki getirmeye calis DONE
+- linki yeni sayfada acmaya calis DONE
+- yaziyi getirmeye calis / limitle 145 karaktere DONE
+- tarihi duzgun calistir DONE
+- medium logosunu ekle DONE
+- paddingleri ekle
```
omonk: This you missed this 😉

SemraSh: hahahah Oh My God this is Turkish🤣

src/components/Common/TitleAndMediaList.js
```diff
@@ -59,7 +59,7 @@ const TitleAndMediaList = ({
       </MediaItems>
       {CTALink && CTAText ? (
         <Padding top={3}>
-          <StyledLink external href={CTALink} title={CTAText}>
+          <StyledLink to={CTALink} title={CTAText}>
```
omonk: Keep in mind this is a Common component and used throughout the site not just for the medium post. I think we need to add props to this component to be able to switch between `external` and `to` props when required.

SemraSh: Thank you for this 🙌

src/components/Common/TitleAndMediaList.js
```diff
@@ -59,7 +60,7 @@ const TitleAndMediaList = ({
       </MediaItems>
       {CTALink && CTAText ? (
         <Padding top={3}>
-          <StyledLink external href={CTALink} title={CTAText}>
+          <StyledLink external={external} href={CTALink} title={CTAText}>
```
omonk: We need to make sure that we can go to either a local page with `to` prop or an external link with `href` prop.

You're setting `external` to false but not providing a `to` prop so you don't end up with a Gatsby `Link` component, just an anchor with an href which will cause issues with how gatsby routing works.

Remember how StyledLink works:
```
if (to) {
    return (
      <Anchor {...externalProps} to={to} {...props}>
        {children}
      </Anchor>
    )
  }

  return (
    <Anchor {...externalProps} as="a" href={href} {...props}>
      {children}
    </Anchor>
  )
```

```
const styledLinkProps = {
    title: CTAText,
    ...(external ? { external: true, href: CTALink }  : { to: CTALink}),
}

<StyledLink {...styledLinkProps}>CTAText</StyledLink>
```

src/components/Blog/MediumPostPreview.js
```diff
@@ -0,0 +1,176 @@
...
+  author,
+  virtuals
+}) => {
+  const previewText =
+    virtuals.subtitle.length < 145
+      ? virtuals.subtitle
+      : virtuals.subtitle.slice(0, 145) + '...'
+  const formattedDate = format(createdAt, 'MMMM DD')
+
+  const imageUrl = `https://cdn-images-1.medium.com/max/2000/${
```
omonk: Screenshot 2019-07-30 at 13 20 41
`/max/2000/` in the URL will request all images at 2000px width which as you can see from the screenshot means they're all > 100kb and will slow down rendering the page for mobile users on slow connections. Because we're not using gatsby-image here there is no sort of 

From a bit of checking I think that the max width these images can go to is 480px so can you replace the url with something like 500px. I appreciate that the images come at different aspect ratios so you might need to do a bit checking to make sure they all take up the 1:1 container we've made, but I'm fairly confident they will!

PR #305, opened 2019-07-23
WIP: Feat/landing page messaging

omonk:
## Description - [Trello ticket](https://trello.com/c/8DoNpMY1/578-landing-page-messaging-update-visuals)

This is the first iteration of the landing page update
- Main changes include updating the homepage landing section
- Extensive changes to the top nav to allow homepage to have a bluebackground. Updates include using data from contentful and page slug to work out if page is service or speciality, removes the need to have a bunch of regex to work out data we already have.


src/components/Header/TopNav/LogoLink.js
```diff
@@ -22,40 +18,45 @@ const StyledLink = styled(Link)`
...
+const HomepageLogo = styled.img`
+  width: 49px;
+  height: 36px;
+  margin-top: ${remcalc(6)};
+`
+
+const LogoLink = ({ slug, isServicePage, isSpecialityPage, isHomePage }) => {
+  const renderSvg = isServicePage || isSpecialityPage || isHomePage
+
+  const fillColorInitial = logoColors[slug] || logoColors['default']
```
GBouffard: `logoColors[slug || 'default']` ; only removes one word but every little helps!

src/components/Header/TopNav/LogoLink.js
```diff
@@ -22,40 +18,45 @@ const StyledLink = styled(Link)`
...
 
-const LogoLink = ({ path = '/' }) => {
-  const isServicePage = getIsServicePage(path)
-  const isSpecialityPage = path.includes('speciality')
+  const fillColorHover = isSpecialityPage
+    ? logoColors.specialityHover
+    : logoColors.defaultHover
```
GBouffard: Same with these 2: fillColorHover and defaultHover:
`const fillColorHover = logoColors[isSpecialityPage ? 'specialityHover' : 'defaultHover']

src/pages/index.js
```diff
@@ -51,17 +58,120 @@ const getHomepageConferences = (events = []) =>
...
+          <Col width={[1, 1, 1, 1, 7 / 12]} style={{ position: 'relative' }}>
+            <IntroSectionTitleWrapper>
+              <SectionTitle reverse as="h1">
+                {introCopy}
+              </SectionTitle>
+            </IntroSectionTitleWrapper>
+            <Subtitle reverse muted>
+              What we offer
+            </Subtitle>
+            <IntroLinkWrapper>
```
GBouffard: an improvement that I think would be nice is to iterate from an array of object
```
const services = [{
  title: 'Software engineering consultancy',
  to: '/engineering'
}]
```
and
```
{services.map(service =>
  <StyledCardTitle noPaddingTop as={Link} reverse to={service.to}>
    {service.name}
  </StyledCardTitle>
```
it will make it easily extendable when we have new services coming.

src/pages/index.js
```diff
@@ -51,17 +58,120 @@ const getHomepageConferences = (events = []) =>
...
+              <StyledCardTitle noPaddingTop as={Link} reverse to="/engineering">
+                Software engineering consultancy
+              </StyledCardTitle>
+              <br />
```
GBouffard: `<br />` . Is it acceptable to use br?

src/components/Header/TopNav/ServiceLink.js
```diff
@@ -21,26 +21,23 @@ const StyledServiceLink = styled(Link)`
...
+  return isServicePage || isSpecialityPage ? (
+    <StyledServiceLink
+      to={`/${service}`}
+      title={service}
+      color={isServicePage ? 'text' : 'white'}
+    >
+      {capitalize(service)}
+    </StyledServiceLink>
+  ) : null
 }
```
GBouffard: I hate a ternary that has `: null` in React. I know it's legacy code but let's improve:
`(isServicePage || isSpecialityPage) && <StyledServiceLink`

src/components/Header/TopNav/LogoLink.js
fabiommmoreira: @GBouffard that way it wouldn't fallback to `default` it `slug` is defined, even if it's not mapped on `logoColors`

src/components/Header/TopNav/LogoLink.js
fabiommmoreira: @GBouffard  same thing here

src/components/Header/TopNav/LogoLink.js
```diff
@@ -22,40 +18,45 @@ const StyledLink = styled(Link)`
...
-          onMouseLeave={() => setFillColor(originalFillColor)}
-        >
-          <ServiceSpecialityLogo
-            fillColor={fillColor}
-            textColor={
-              logoColors[isServicePage ? 'defaultText' : 'specialityText']
-            }
-          />
-        </StyledLink>
+      {renderSvg ? (
```
fabiommmoreira: I think it's more readable to have an `if` instead of nested ternary operators

```
if(renderSvg) {
    return (
        isServicePage || isSpecialityPage ? (
          <StyledLink
            to="/"
            title="Return to Homepage"
            onMouseEnter={() => setFillColor(fillColorHover)}
            onMouseLeave={() => setFillColor(fillColorInitial)}
          >
            <ServiceSpecialityLogo
              fillColor={fillColor}
              textColor={textColor}
            />
          </StyledLink>
        ) :
          <HomepageLogo src={HomepageSvg} alt="yld logo" />
    )
}

return (
    <Link to="/" title="Return to Homepage">
        <img role="link" height="48" src={logo} alt="yld logo" />
    </Link>
)
```

src/components/Header/TopNav/ServiceLink.js
```diff
@@ -21,26 +21,23 @@ const StyledServiceLink = styled(Link)`
...
-          to={`/${service}`}
-          title={service}
-          color={isServicePage ? 'text' : 'white'}
-        >
-          {capitalize(service)}
-        </StyledServiceLink>
-      ) : null}
-    </Fragment>
+const getServiceFromSlug = slug =>
+  Object.keys(specialitiesMap).find(speciality =>
```
fabiommmoreira: We're already exporting an `Object.keys` of `specialitiesMap` on `navLinksHelper.js` under `servicesList`, you can use that on instead ;)

src/components/Header/TopNav/ServiceLink.js
fabiommmoreira: I agree with @GBouffard ;)

src/components/Header/TopNav/index.js
```diff
@@ -34,12 +35,56 @@ const TopNavList = styled.ul`
   }
 `
 
-const TopNav = ({ links, themeVariation, path }) => (
+const getSlugs = (arr = []) => arr.map(({ slug }) => slug).filter(i => i)
```
fabiommmoreira: `const getSlugs = (arr = []) => arr.filter(({ slug }) => slug)`

No need to map before filter ;)

src/components/Header/TopNav/ServiceLink.js
omonk: true but I'm also using the `specialitiesMap` value within the map to reference the current key, `specialitiesMap[speciality]`. I think it's much safer to use this here rather than importing the list and the map to generate this service from slug. 

Exporting `const a  = Object.keys(b);` seems on the whole fairly unnecessary as well

src/pages/index.js
omonk: I believe so yes!

src/pages/index.js
omonk: This is now in contentful so am doing that!

src/components/Header/TopNav/index.js
omonk: @fabiommmoreira I removed to just be filter but I actually wanted to get the slugs then filter the array, not filter the array based on whether the object at current index had valid slugs.

src/components/Header/TopNav/LogoLink.js
GBouffard: Good point. I didn't think of this part. Thanks Fabio

src/components/Header/TopNav/LogoLink.js
```diff
@@ -23,52 +18,63 @@ const StyledLink = styled(Link)`
...
-          />
-        </StyledLink>
-      ) : (
-        <Link to="/" title="Return to Homepage">
-          <img role="link" height="48" src={logo} alt="yld logo" />
-        </Link>
-      )}
-    </Fragment>
+    <Link to="/" title="Return to Homepage">
+      <img role="link" height="48" src={logo} alt="yld logo" />
```
AniaMakes: I'm wondering how this is read by screen-readers. Mostly wondering if the alt should be yld logo - navigate to home, because if it's read as an image rather than link then that cuts navigation out.

PR #307, opened 2019-07-23
Case studies order

GBouffard:
## Description - [Ordering case studies on the site](https://trello.com/c/0nR4iqbo/561-ordering-case-studies-on-the-site-how-do-we-do-it)

- order agreed: V2, allContentfulTemplatedCaseStudy, allContentfulNonTemplatedCaseStudy.
- Order fully customisable via Contentful `ourWork`

## Review template


src/pages/our-work.js
```diff
@@ -13,6 +13,22 @@ import Hr from '../components/Common/Hr'
...
+  'doctorlink',
+  'central-working',
+  'learnerbly',
+  'trainline',
+  'the-economist',
+  'kingfisher',
+  'Thomas-Cook',
+  'joyent',
+  'canon'
+]
```
omonk: I thought we agreed that the sorting shouldn't be in the code? The ordering should be controlled by contentful and the code should handle fallbacks.

GBouffard: I thought it would make sense too. It's ideal but I didn't see it in the ticket so worked in the code. If we agree on something like that it should be in the ticket description.

The Contentful API gives us `publishedAt`, `createdAt` and `updatedAt` which I looked at and didn't manage to make it work. I can look into it again.

src/pages/our-work.js
```diff
@@ -41,18 +30,42 @@ const ourWork = {
...
-    allContentfulNonTemplatedCaseStudyV2
+    contentfulOurWork: { caseStudies }
   } = data
 
+  const displayOrderByIDs = caseStudies
+    .filter(({ publish }) => publish)
+    .map(({ id }) => id)
+
+  const getCSOrderIndex = id =>
+    displayOrderByIDs.indexOf(displayOrderByIDs.find(cs => id.includes(cs)))
```
omonk: You don't need the find in this instance. If indexOf can't find the item in the array it will return `-1`, if the find method can't find it it just returns undefined so if the id is not in the array you're doing `ids.indexOf(undefined)` if you include the `.find` here.

GBouffard: Wow! So true. How come I left this?? Updated. Now gonna have a look at your fiddle!

src/pages/our-work.js
```diff
@@ -54,17 +54,26 @@ const OurWork = ({ data }) => {
...
+  const mappedFromContentfulOrder = displayOrderByIDs.map(orderedId =>
+    allCaseStudies.find(cs => cs.id === orderedId)
+  )
+  const missingFromContentfulOrder = allCaseStudies.filter(
+    cs => !displayOrderByIDs.includes(cs.id)
+  )
+  const orderedCaseStudies = [
+    ...mappedFromContentfulOrder,
+    ...missingFromContentfulOrder
+  ]
```
omonk: Nice, this makes it much easier to read and reason with!

src/pages/our-work.js
```diff
@@ -92,7 +101,7 @@ const OurWork = ({ data }) => {
       </Grid>
       <GreyBackground>
         <Grid>
-          {allCaseStudies.map((caseStudy, index, arr) => {
+          {orderedCaseStudies.map((caseStudy, index, arr) => {
```
omonk: I would `orderedCaseStudies && orderedCaseStudies > 0 && ....` here just for peace of mind

PR #309, opened 2019-07-25
Update readme with theme, Grid Row and Col explanations

GBouffard:
## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,53 @@ The aim of this lammbda is to have up to date metrics of yld's open source contr
 
 Main [`@yld.io/oss-stats`](https://www.npmjs.com/package/@yldio/oss-stats) to aggregate all open source contributions for `yldio` organization members.
 
+## Styling
+
+### theme
+
+The whole application is wrapped in a theme file `src/utils/theme.js` that contains the official breakpoints, spacing and colors that can be used within the app. It's always available within a styled component as `theme` prop.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,6 +125,53 @@ The aim of this lammbda is to have up to date metrics of yld's open source contr
...
 
+## Styling
+
+### theme
+
+The whole application is wrapped in a theme file `src/utils/theme.js` that contains the official breakpoints, spacing and colors that can be used within the app. It's always available within a styled component as `theme` prop.
+
+### Grid
+
+We often use a system of Grid, Row and Column; all defined within `src/components/grid.js` to style sections. They can be confusing at first. The Grid component is a div that automatically defines how a section behaves accoding to the official breakpoints of the app.
```
omonk: This is a bit misleading. The site is designed using a grid system

GBouffard: Updated

omonk:
The additions here are more of an explanation of a CSS responsive grid which is pretty self evident in itself. You'll see this system being used in many many projects you work on throughout your career. I really hope I'm not simplifying this because of previous knowledge but the explanation here I think doesn't do the theory of a grid system justice. This is what the meeting I've called is all about.

I think you're able to slim it down to say we use a standard responsive 12 column CSS responsive grid.

I will find a suitable link to include here

PR #312, opened 2019-07-26
Make section headers on the landing page clickable

GBouffard:
## Description - [ticket](https://trello.com/c/7cJecfDy/566-make-section-headers-on-the-landing-page-clickable)

made the titles as links if the page exists

## Checklist


omonk:
ah yes this is because open source is not a service. I believe all you need to do is extend the open source content type to include a pageReady boolean and a slug, then you should be good 👌

PR #315, opened 2019-07-29
Case study - adjust space between video and text

fabiommmoreira:
## Engineering case studies inconsitencies - [508](https://trello.com/c/V0FkdBU5/508-engineering-case-studies-inconsitencies)

Only had to fix "Address spacing between initial paragraph and the testimonial video - should be 72px".

Other points seemed to be already fixed.

## Checklist


src/templates/caseStudy.js
```diff
@@ -36,11 +36,13 @@ const CaseStudy = ({
...
       <GreyBackgroundOffset topMargin topOffset={-150}>
         <Grid>
-          <VideoSection src={videoInfo[0]} />
           <Padding
             top={{ smallPhone: 3, tablet: 4 }}
-            bottom={{ smallPhone: 3.5, tablet: 4 }}
+            bottom={{ smallPhone: 3, tablet: 4 }}
           >
+            <VideoSection src={videoInfo[0]} />
+          </Padding>
```
omonk: We can do a lot of improvements to this section here. Specifically the whole HTML structure looks a bit suspect here.

omonk: Moving the `VideoSection` into the `Padding` removes the `Grid > Row > Col`  markup we want. I would opt for putting the top padding on the VideoSection and the bottom padding on the  `SeconTextSection`. 

To be honest I'd even argue this components (FirstTextSection, SecondTextSection) don't even need to be broken out into individual components. This separation isn't really necessary and just means that the HTML markup we want with the grid system is obscured.

fabiommmoreira: @omonk you're right, we definitely should avoid breaking the `Grid > Row > Col` structure. I didn't notice VideoSection was an actual Row container. PR updated to remove this `<Padding>` components from between grid sub-component layers ;)

PR #321, opened 2019-07-31
HOTFIX: add react-native to engineering services

omonk:
## Description - Hotfix

For some reason react native displays design, this logic has been removed in the landing page updates so this is a very short lasting change 

https://www.yld.io/speciality/react-native/


omonk:
Will be sorting that all out so its automated meaning we don't have to add them manually every time

PR #322, opened 2019-07-31
Unit tests

fabiommmoreira:
## [Unit tests](https://trello.com/c/PixUazAK/588-unit-tests-spike)

Configuration to test with react-testing-library
- create separate jest config file to have different testMatch, coverageThreshold and setupFiles
- add script to package.json
- add step to circleci
- update gitignore to disregard coverage folder
- write interaction tests for AboutUs/Teams
- write stories for the rest of About Us components

## Checklist


package.json
```diff
@@ -60,6 +60,7 @@
     "test": "run-s test:**",
     "test:snapshots": "jest --testPathPattern=\"storyshots.test.js\" --silent",
     "test:snapshots:update": "jest --testPathPattern=\"storyshots.test.js\" -u --silent",
+    "test:interaction": "jest --config jest-unit.config.js",
```
omonk: Is it worth naming this config something else as we have a script of `test:unit` which just runs unit tests for the lambdas currently, https://github.com/yldio/yld.io/pull/322/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R66

src/components/AboutUs/StaffCard.js
```diff
@@ -58,9 +58,16 @@ const SocialLinks = ({ data }) => {
...
+  name,
+  image,
+  role,
+  description,
+  socialLinks = [],
+  ...props
+}) => {
   return (
-    <Col width={[1, 1, 1, 1, 6 / 12, 4 / 12]}>
+    <Col width={[1, 1, 1, 1, 6 / 12, 4 / 12]} {...props}>
```
omonk: Can we add the `data-testid` explicitly instead of spreading the rest of the props, seems a bit dangerous

package.json
fabiommmoreira: Right, I thought the same but then forgot to actually change the file name 😛

.circleci/config.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - checkout
       - *attach_workspace
       - run:
-          command: npm run test:snapshots
+          command: npm run test:snapshots && npm run test:interaction
```
omonk: Not a change request, just discussion. If we use yarn throughout the project should we use yarn for all of our CI commands as well? I don't think it should make too much of a difference as modules are installed with yarn but just a though.

PR #327, opened 2019-08-02
isolate storyshots into separate snapshots

fabiommmoreira:
## Create separate snapshots from storyshots - [588](https://trello.com/c/PixUazAK/588-unit-tests-spike)

Previously we had our storyshots generating a single snapshot. That make it hard to which sections were actually impacted by changes in a new commit. Having separate snapshots shoud make it easier and avoid doing blind snapshot updates.
## Checklist


AniaMakes:
There is something wrong here... It was failing on snapshots, so I updated the snapshots. Running the update locally again still generated new files. Pushed that. I'm back to square one :-/

Logs from my terminal : 
image
image
image

PR #329, opened 2019-08-05
Feat/lever test 

AniaMakes:
## Description - [Trello ticket 557](https://trello.com/c/vhmP81Ee/557-lever-function-test)

Added tests for lever lambda

## Checklist


package.json
```diff
@@ -61,6 +61,9 @@
     "test:snapshots:update": "jest --testPathPattern=\"storyshots.test.js\" -u --silent",
     "test:e2e": "npm run build:app && testcafe chrome:headless ./tests/e2e/**/*.js -S -s artifacts/screenshots --reporter html:artifacts/screenshots/report.html --assertion-timeout 10000 --selector-timeout 500000",
     "test:lighthouse": "npm run build:app && jest --testPathPattern=\"lighthouse.test.js\"",
+    "test:lambdas": "run-s test:lambda:**",
```
GBouffard: Do we still need `test:lambda:leverDifferent` and `test:lambda:leverSame` ?

tests/lambdas/leverDifferent.test.js
```diff
@@ -0,0 +1,136 @@
...
+    got.post.mockReturnValueOnce({ body: leverBody })
+
+    response = await lambda.handler({
+      headers: {
+        authorization: 'basic foo_test_thingy'
+      }
+    })
+  })
+
+  it('Calls the lambda leaver webhook', () => {
```
GBouffard: Typo: `Lever` ;)

tests/lambdas/leverDifferent.test.js
```diff
@@ -0,0 +1,136 @@
...
+    expect(got.post).toHaveBeenCalledTimes(1)
+  })
+
+  it('returns the response body from the lever webhook', () => {
+    expect(response.body).toEqual(leverBody)
```
GBouffard: I think it's fine. I'm still trying to understand how the whole Lever lambda works.
One thing which I think is missing is testing the 2 other cases for `return`. According to what I understand the lambda can return 3 different things:
- the new list of jobs in prod if different than what we had.
- 'Difference in jobs found but this is not production so no deployment for you' if different than what we had but environment is prod.
- 'No differences in jobs found, not deploying' if no difference.
I think the 3 should be tested to make sure than the lambda works as expected. What do you think Anna? I would be keen to get other opinions from the engineers in the team on this. @Couto @SemraSh ?

AniaMakes: Fair point about testing for case 2 in your example.

Couto: The convention is to just use one file indeed, unless the code/scenario is really complex, which is not the case.

We are testing a single function so we should merge those two files into a single `lever.test.js` file, where we can have `describe` blocks one for each scenario.

I agree with @GBouffard here, we have three possible outcomes:
- Returned jobs are equal. Nothing to do.
- Returned jobs are different:
  - is production: is the return value expected?
  - is not production: is the return value expected?

So it seems to me the last case is missing (I might be wrong)

AniaMakes: So I rejigged this a little bit - we still have two separate files, but now one is Prod and the other is Dev. Basically the function needs a isProd variable, so that needs to be passed as true / false, so it's easier to have two separate files with different scripts. If you have any suggestions how to make it work in a single file, I'm all ears

omonk:
@AniaMakes I have made quite a few changes to the lever tests here, please take a look: https://github.com/yldio/yld.io/compare/feat/lever-unit-tests?expand=1

tests/lambdas/leverDev.test.js
```diff
@@ -0,0 +1,79 @@
...
+      protocol: 'https:',
+      username: '',
+      password: '',
+      host: 'api.lever.co',
+      hostname: 'api.lever.co',
+      port: '',
+      pathname: '/v0/postings/yld',
+      search: '?mode=json',
+      hash: ''
+    })
```
omonk: Here we don't need to mock the URL module, its a native JS module so we can rely on it to give us what we want. 
Using this mock here means that we aren't making sure we're passing the URL module the correct values, because we're in control of what we pass it, it's important to make sure we test it. 

In my version you can see we're testing that the got module is called with the correct object from the URL as the first argument. The reason we do it like this is that if there are changes in the lever function to the got requests and urls then the tests will fail. 

[We set up the URLS](https://github.com/yldio/yld.io/compare/feat/lever-unit-tests?expand=1#diff-4772b57a44b209349e50cff83b53d1c8R17-R18)
[then we make sure that the got mock is called with the right arguments](https://github.com/yldio/yld.io/compare/feat/lever-unit-tests?expand=1#diff-4772b57a44b209349e50cff83b53d1c8R59-R62)

tests/lambdas/leverDev.test.js
```diff
@@ -0,0 +1,79 @@
...
+const got = require('got')
+const url = require('url')
+
+jest.mock('got')
+jest.mock('url')
+
+describe('different data, not production', () => {
+  let response
+
+  beforeAll(async () => {
```
omonk: Usually `beforeAll` blocks are used to set up/take down/reset mocks and other functionalities that the test needs to function. It is possible to but not usually used to perform the execution of the function at hand.

tests/lambdas/leverDev.test.js
```diff
@@ -0,0 +1,79 @@
+const lambda = require('../../src/functions/lever')
+const got = require('got')
+const url = require('url')
+
+jest.mock('got')
+jest.mock('url')
+
+describe('different data, not production', () => {
```
omonk: Following on from [this](https://github.com/yldio/yld.io/pull/329/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.yml#r313721201) comment. It is worth taking a look at the documentation from jest to understand how describe blocks are intended to be used, please read these two links: 

- [Describe](https://jestjs.io/docs/en/api#describename-fn)
- [Test](https://jestjs.io/docs/en/api#testname-fn-timeout) - this is an alias of `it` which we are using in our tests. 

What I've created in my branch is something that looks like this:
```
describe(lever lambda)
    afterEach(() => {
     jest.clearAllMocks()
   })
    it('should call the lever webhook and return X when there are differences', () => { // test // })
    it('should only return X and not call lever webhook when there are no differences', () =>  { // test // })
```

This set up means that we have a much better way of composing each unit test for our function that we want to test. Mocks are set in the individual unit test (`it(...)`) and then reset after each test has finished running.

Instead of running multiple describe blocks for the same function we're isolating it to one block and making it a lot more concise. Another benefit of this is understanding the output from lever when we are writing tests and also looking at failed test reports in CI

```
PASS  tests/lambdas/lever.test.js
  Lever
    ✓ fires a request to the LAMBDA_LEVER_WEBHOOK url when there are differences in jobs (6ms)
    ✓ returns 200 and no difference in body (1ms)
````

Whereas the set up in this branch would be

```
FAIL tests/lambdas/leverDev.test.js
   different data, not production
     ✕ returns 200 and no difference in body
```

File: tests/lambdas/leverDev.test.js [hunk 7]
```diff
@@ -0,0 +1,79 @@
...
+        {
+          id: 'five'
+        }
+      ]
+    })
+
+    response = await lambda.handler({
+      headers: {
+        authorization: 'basic foo_test_thingy'
+      }
```
omonk: To remove the need to run these tests with hardcoded auth tokens and passing a header here we can use jest mocking to get around all of this by mocking the `Auth` module, I've done this [here](https://github.com/yldio/yld.io/compare/feat/lever-unit-tests?expand=1#diff-4772b57a44b209349e50cff83b53d1c8R1)

An  overview of what the auth and lambdas are doing.
```
const leverCallback = () => { // do the lambda stuff here }

const Auth = (event, callback) => {
    // do stuff with event data
   if (event data is ok) {
       return callback()  <---- remember this is is just the leverCallback in this instance
   }
}

const lambda = (event) => Auth(event, leverCallback)
```

With the line
```
jest.mock('../../src/functions/utils/auth', () => jest.fn((_, cb) => cb()))
```

All this says is, "ignore the first argument, just return the callback function and execute it", so we bypass the `Auth` module and just fire the lambdaCallback

AniaMakes: Yeah, I figured it's just going "eh, just give me the callback", but is there a technical reason why doing this is better? From my perspective it's doing the same thing, just a bit differently...

tests/lambdas/leverDev.test.js
AniaMakes: So at which point is it "safe" to just use the modules as-is? You said "native JS module, so we can rely on it". Would you say the same about, for example, lodash? Or "any module that just transforms data (as opposed to fetching / sending) is safe to use as-is"? 

> Using this mock here means that we aren't making sure we're passing the URL module the correct values, because we're in control of what we pass it, it's important to make sure we test it.

But isn't mocking about faking the data we want to receive? I'm also having trouble parsing the above sentence - do all the instances of "it" apply to the "URL module".

tests/lambdas/leverDev.test.js
AniaMakes: This tests pass both locally and on CI. The only reason that that I have two different files is because I couldn't figure out how to pass different env.vars for faking prod vs faking dev. 

My bad though because I just copied-pasted the `it` description for the leverDev, and the test is actually testing what it's meant to test.
image

AniaMakes: The way you've done it has 4 `expect` tests in one `it` - I think we're thinking in different levels of testing - to me, it feels each `expect`  be inside a separate `it` so that each gets a separate tick as the tests run. 

Would you always test in pairs, so to speak: that `got` was called with the values we expect, and that the `response` is what we expect it to be? If not, what is the deciding factor as to what gets `expected`?

AniaMakes:
@omonk thank you for all the explanations above - I left comments where I have further questions. I think a lot of the stuff does relate to experience... I have a few additional questions below. 

1. Is there a guidance on how much data you fake? In your `no diff` example you have one item in each array, so in the `diff` version would two / one item suffice?
2. Other than "it's shorter", why is it better to fake a body which is one-word as you did, as opposed to an example the body we expect? I feel that the latter is more helpful long term as it shows the structure of the data we were expecting at the time the function and test were written. 
3. Why you're doing `afterEach clearAllMocks` if each mock has `mockReturnValueOnce`? Feels like `clearAllMocks` is there if your tests have the same calls but different criteria...? Or is it just "that way we know we have a clean slate each time"?
4. Your tests only tests prod stuff. Why have you skipped a test for "not prod, so no deploy for you"?

tests/lambdas/leverDev.test.js
omonk: No lodash is just a library of scripts whereas URL is a native module to JS. We can trust the native module more than something like lodash, although in this unit test we're not mocking the sort function of lodash.

Mocking is about isolating the object under test and making sure that the internals of that function perform the tasks we want them to. The key point here is `isolation`. If you mock the dependencies you know the test is only testing the internals, not externals. Another way of testing this lambda is to use `nock` to return the correct data from an HTTP request, but that means we're still relying on Got to do the request and formulate the http request correctly, which is not what we want to test. 

Mocking in unit testing allows us to control all of the external dependencies of a function. E.g. the got module or even a complex data transformation function that we have written ourselves. Remember, we don't want to test that functionality, we only want to test our module.

If you've not heard of the testing pyramid I suggest reading up on it, Martin Fowler originated the idea I believe: 
image

tests/lambdas/leverDev.test.js
omonk: Imagine we have 100 lambdas and we write tests for all of them and we pass the header you have here: 

```
headers: {
    authorization: 'basic foo_test_thingy'
}
```
Every single lambda test is testing our lambda functionality, great, but it's also testing the Auth function. So what's really happening is that we're writing 100 tests for lambdas and all of them test the Auth function. (This isn't want unit tests are about, thats for integration or e2e tests, something we've not touched on yet. It's really important you read up on the concepts of units tests.)

Anyway, imagine we change the internals of the Auth function to use a `Bearer` token or something changes which means the authorization header here is no longer working. What happens? All 100 tests fail and we have to change *all* of the lambda tests to reflect this change.

The difference between `passing the authorization header to the handler` and `mocking the Auth function` is that with the mock we are positive that the only  thing we're concerned about is the lambda under test, not the Auth function. It's really important to understand the difference here!

tests/lambdas/leverDev.test.js
omonk: Each `it` block is a scenario you want to test, in my branch we're testing two scenarios:

- `fires a request to the LAMBDA_LEVER_WEBHOOK url when there are differences in jobs`
- `returns 200 and no difference in body`

We set up the mocks return values and other various requirements then we execute the function. We want to make sure that our function operates correctly and to make sure a function fires correctly we have to check several things. It's not against the rules to use multiple `expect`s in each testing block. One `it` is a **scenario**, for that scenario to pass we have to assert something or somethings. For a simple function that can be one thing but in this case we need to make sure that we're doing several things correctly.

omonk: What do you mean testing in pairs? A very common pattern in unit tests is `response` (from the function under test) and `expected` (what we expect the function to return). 

The deciding factor of what to test should be fairly self evident if you read through the function you want to test. In the lever function we want to make sure that:
- The URLs are parsed correctly, to do this we `expect got to be called with X` for each url
- If there are difference we want got.post to be called and with the right argument, to do this we `expect got.post to be called with X`
- Then we want to make sure the responses are correct so we just set up `expected` and compare this with the return value of the function.

It's important to understand the core aim of the function, here it's "make sure that the lever webhook url is requested if there are differences in job IDs" and "this is how this happens". Understanding this is then how we decide what to assert on with `expect`

omonk: re prod and dev, i'm not too concerned about the dev instance as it only occurs locally, thus if there are issues when developing them you'll be aware and then need to fix them.

tests/lambdas/leverDev.test.js
omonk: And of course, we will be writing tests for the Auth function!

omonk:
@AniaMakes 

```Is there a guidance on how much data you fake? In your no diff example you have one item in each array, so in the diff version would two / one item suffice?
```
Yeah that should be fine, I'm not sure how much you'd gain from putting say 100 items in each mock return value.

```
Other than "it's shorter", why is it better to fake a body which is one-word as you did, as opposed to an example the body we expect? I feel that the latter is more helpful long term as it shows the structure of the data we were expecting at the time the function and test were written.
```
Do you mean the `got.post` return body we're mocking? If yes then the response from the lever webhook is just a string I believe? The point of the test here is not what gets returned from the got.post request. We're testing our function, so this test is making sure the response from got.post is just returned without any changes. This is a really really core part of unit tests that does take a bit of time to wrap your head around. We have very little concern of what the `got.post` function returns, it's all about what **we** do with that return value, thats what this test is about.

If no, can you expand on where you're referring to please.


```
Why you're doing afterEach clearAllMocks if each mock has mockReturnValueOnce? Feels like clearAllMocks is there if your tests have the same calls but different criteria...? Or is it just "that way we know we have a clean slate each time"?
```

Yes it's more of a clean slate thing than anything else, as you said we're using `mockReturnValueOnce`. I believe that if we removed the `mockReturnValueOnce` in the second `it` block in my example that either we'd get undefined and the test would fail or it would try and make a request to the urls we give it, not sure. 

Generally it's just good practise to make sure each `it` block is starting from a fresh and not affected by other scenarios we're testing.

```
Your tests only tests prod stuff. Why have you skipped a test for "not prod, so no deploy for you"?```

PR #332, opened 2019-08-07
Change the opacity of the Learnerbly copy

SemraSh:
## Trello ticket - [619](https://trello.com/c/1yNEkwC7/619-speciality-amend-lernerbly-text-opacity)

Changing the opacity of a case study.


src/components/Speciality/Projects.js
```diff
@@ -11,31 +11,38 @@ const Emphasis = styled.em`
...
-          alt={project.posterImage.title}
-          src={project.posterImage.file.url}
-          style={{ maxHeight: '100%' }}
-        />
-      </PosterImage>
-    </section>
-  </AnimatedLink>
-)
+const PosterLinks = ({ project }) => {
+  const backgroundIsDark = project.posterColor === 'ff4e5b' // Learnably poster color
```
GBouffard: I would rename to `isDarkBackground`. It's better to have a boolean variable start with "is".
By the way is this working? In theory it looks good but when I go to `http://localhost:8000/speciality/data-analysis/` the text is still dark. Is this my local environment?

PR #336, opened 2019-08-13
remove twitter:img and twitter:card tag from head

omonk:
## Description - [Trello ticket](https://trello.com/c/G18faAEA/378-fix-hero-images-when-posting-links-of-the-website-on-social-media)

Previously this was working and now twitter doesn't seem to want to pick up the images for case studies... removing these tags as og tags should  suffice.

These two tags fallback to their corresponding `og` tags so we can remove them.

### DIFF
--- a/src/components/Common/Head.js
+++ b/src/components/Common/Head.js
@@ -51,9 +51,7 @@ const Head = ({ page }) => {
             <meta property="og:url" content={siteUrl} />
 
             {/* Twitter */}
-            <meta name="twitter:card" content="summary" />
             <meta name="twitter:site" content="yldio" />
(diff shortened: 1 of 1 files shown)


omonk:
@AniaMakes - yes this is something I'm still not sure about. I've seen on twitters docs that they have URLs with just `//` and not `https:`.

The twitter sharing did work previously with the same set up but I'm not sure why now it's  not...

omonk:
Or maybe this is what broke it: https://github.com/yldio/yld.io/pull/317/files#diff-bef7bc103fc66c07819d10e073b2df74L38

To add this back in I believe we need to check if the url starts with https first. I will look into it

PR #337, opened 2019-08-13
CI updates - Build first then run tests

omonk:
## Description

This is an update to all of our CI processes, first we build then test, previously it was building the gatsby app on every step which seems fairly pointless.

This should help with some issues we're seeing regarding steps fail due to not getting the right cache.


.circleci/config.yml
```diff
@@ -1,72 +1,79 @@
...
 version: 2
 workflows:
   build_and_test:
     jobs:
-      - lint
-      - lighthouse
-      - snapshots
-      - testcafe
+      - build
+      - lint:
```
AniaMakes: I understand adding build to all the commands that had `npm run build` as first part of the command. But why does lint need it if it doesn't need the build to happen to work?

package.json
```diff
@@ -59,8 +59,8 @@
     "test": "run-s test:**",
     "test:snapshots": "jest --testPathPattern=\"storyshots.test.js\" --silent",
     "test:snapshots:update": "jest --testPathPattern=\"storyshots.test.js\" -u --silent",
-    "test:e2e": "npm run build:app && testcafe chrome:headless ./tests/e2e/**/*.js -S -s artifacts/screenshots --reporter html:artifacts/screenshots/report.html --assertion-timeout 10000 --selector-timeout 500000",
```
GBouffard: I remember that we had to add `npm run build:app &&` in both for local reasons. Without it it was randomly failing. I worked with Jen on that. Did you check that locally it doesn't fail?

omonk: For local development you will need to have a local build of the app yeah. I didn't check locally but as we point to `public/` in the e2e createServer method for this testing suite this is quite straight forward.

I will add something to the readme regarding having to have a local build

.circleci/config.yml
omonk: Actually we do need this as we use the node modules from build for lint

PR #343, opened 2019-08-15
606 how we work speciality

SemraSh:
## Description

Trello tickets 
[606](https://trello.com/c/8Y3zWDAM/606-add-how-we-work-with-speciality-section-to-speciality-page)
[618](https://trello.com/c/9m5bZg7p/618-central-working-case-study-card-updates)
[619](https://trello.com/c/1yNEkwC7/619-speciality-amend-lernerbly-text-opacity)

- Adds new section to Speciality template `HowWeWorkSection`
- Adds to query
- Updates on designs of PosterLinks, YLD logo background color


src/components/Speciality/Projects.js
```diff
@@ -4,38 +4,60 @@ import { Row, Col, Grid } from '../grid'
...
-          style={{ maxHeight: '100%' }}
-        />
-      </PosterImage>
-    </section>
-  </AnimatedLink>
-)
+const PosterLinks = ({ project }) => {
+  const preferAdditionalPoster =
+    project.slug === 'central-working-changing-how-you-work'
+
```
omonk: To allow us to not have to update this code each time we want to add an alternative poster image I would check for the existence of additionalPoster in the project object instead.

```
const preferAdditionalPoster = project && project. additionalPoster ! == null
```

omonk: If you wanted to be really picky you could do: 

```
const imageKey = project && project. additionalPoster ! == null ? 'additionalPoster' : 'posterImage'

const poster = {
    title: project[imageKey].title
    url: project[imageKey].file.url,
    color: project.posterColor
  }
```

src/components/Speciality/Projects.js
```diff
@@ -4,38 +4,60 @@ import { Row, Col, Grid } from '../grid'
...
+    title: preferAdditionalPoster
+      ? project.additionalPoster.title
+      : project.posterImage.title,
+    url: preferAdditionalPoster
+      ? project.additionalPoster.file.url
+      : project.posterImage.file.url,
+    color: project.posterColor
+  }
+
+  const isDarkPosterColor = getColorLuminance(poster.color) < 132.5
```
omonk: Is there a reason for choosing `132.5` instead of `127.5`?

src/templates/speciality.js
```diff
@@ -122,6 +122,12 @@ export const pageQuery = graphql`
               url
             }
           }
+          additionalPoster {
+            title
+            file {
+              url
+            }
+          }
```
omonk: Can you have a go at updating this and the rest of the poster images to use the contentful fluid queries? Then where we use it we'll need to use common `Image` component

src/components/Speciality/Projects.js
SemraSh: But we don't know if we will always prefer the additional poster for the Projects, additional poster might not be null in the cases we don't want to use it here.

src/components/Speciality/Projects.js
SemraSh: I didn't know 127.5 refers to a specific lightness, 132.4 is Learnerbly luminance. Is 127.5 a special value in terms of luminance?

SemraSh: central-working luminance:  `243`
learnerbly luminance: `132.4`

both above `127.5` so I've changed the limit to be 132.5 closer to `learnerbly`.

src/components/Speciality/Projects.js
omonk: But we will only ever put the image in the field in the CMS if we want it right?

src/components/Speciality/Projects.js
```diff
@@ -4,38 +4,58 @@ import { Row, Col, Grid } from '../grid'
...
+          >
+            {project.introSentence.introSentence}
+          </BodyPrimary>
+        </CardHeader>
+        <PosterImage justifyCenter alignCenter color={poster.color}>
+          <img
+            alt={poster.title}
+            src={poster.url}
+            style={{ maxHeight: '100%' }}
+          />
```
omonk: This should be the `Image` component from Common components. Same as [here](https://github.com/yldio/yld.io/blob/6fd316cf5d5747360b9c45d5599a5612189aa993/src/components/AboutUs/Subsidiaries.js#L14) but without the `width` music.

Instead of grabbing all the image data in `poster` and passing each of the keys down as prop values you can just do:

`<Image image={project[imageKey]} />`

This should do everything you need!

src/components/Speciality/HowWeWorkWith.js
```diff
@@ -0,0 +1,54 @@
...
+    padding-bottom: ${({ theme }) => theme.space[7]};
+  `}
+`
+
+const HowWeWorkWithSection = ({
+  howWeWorkWithTitle,
+  howWeWorkWithCopy,
+  howWeWorkWithPractises
+}) => {
+  return (
```
GBouffard: Just a small detail but the `return` does not need to be there. `=> (` automatically returns

src/components/Speciality/HowWeWorkWith.js
```diff
@@ -0,0 +1,54 @@
...
+    <GreyBackground>
+      <Grid>
+        <StyledRow style={{ justifyContent: 'space-between' }}>
+          <Col width={[1, 1, 1, 1, 6 / 12, 6 / 12, 5 / 12]}>
+            <SectionTitle>{howWeWorkWithTitle}</SectionTitle>
+          </Col>
+          <Col width={[1, 1, 1, 1, 6 / 12, 6 / 12, 5 / 12]}>
+            <BodyPrimary>{howWeWorkWithCopy}</BodyPrimary>
+            {howWeWorkWithPractises &&
+              howWeWorkWithPractises.length > 0 &&
```
GBouffard: I'm sure we already talked about it but I guess `&& howWeWorkWithPractises.length &&` didn't work?

PR #350, opened 2019-08-19
Github lambda - Repos util test

GBouffard:
## Description - [Github lambda - Repos util test](https://trello.com/c/6T3PNL74/633-repos-and-meta-lambda-tests)

Put a short summary of what you did here:
Lambda testing fun!!

## Checklist


tests/unit/lambdas/repos.test.js
```diff
@@ -0,0 +1,117 @@
...
+const sameRepoOne = {
+  url: 'https://github.com/yldio/fake-repo',
+  nameWithOwner: 'yldio/fake-repo',
+  descriptionHTML: '<div>This is a fake repo.</div>',
+  starCount: 10924,
+  pullRequestCount: 4
+}
+
+const differentRepoOne = Object.assign({}, sameRepoOne)
+differentRepoOne.pullRequestCount = 6
```
omonk: This is the same thing but keeps it a bit neater.
```
const differentRepoOne = {
    ...sameRepoOne,
    pullRequestCount: 6
}
```

tests/unit/lambdas/repos.test.js
```diff
@@ -0,0 +1,117 @@
...
+  })
+
+  it('should not call updateEntry and no changes should be returned if repos and contentfulRepos have no differences', async () => {
+    const response = await ReposLambda(mockedEnvironment, {
+      repos: [sameRepoOne, sameRepoTwo]
+    })
+
+    expect(ossUtils.updateEntry).not.toHaveBeenCalled()
+
+    const emptyAcc = []
```
omonk: can we name this `expected` instead 

`expect(response).toStrictEqual(expected)`

tests/unit/lambdas/repos.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+    expect(ossUtils.updateEntry).toHaveBeenCalledWith(
+      contentfulRepoOne,
+      ossUtils.generateContentfulData(differentRepoOne, repoKeys),
+      mockedEnvironment,
+      differentRepoOne.nameWithOwner
+    )
+
+    const changedReposAcc = [differentRepoOne.nameWithOwner]
+    expect(response).toStrictEqual(changedReposAcc)
```
omonk: Same one as here https://github.com/yldio/yld.io/pull/350/files#r316034401

tests/unit/lambdas/repos.test.js
```diff
@@ -0,0 +1,117 @@
...
+      'starCount'
+    ]
+
+    const response = await ReposLambda(mockedEnvironment, {
+      repos: [differentRepoOne, sameRepoTwo]
+    })
+
+    expect(ossUtils.updateEntry).toHaveBeenCalledWith(
+      contentfulRepoOne,
+      ossUtils.generateContentfulData(differentRepoOne, repoKeys),
```
omonk: Where we do this can we instead write out the object we expect instead? Having the function under test and the test depending on the same function can be an issue.

Say the `generateContentfulData` function changes, e.g. whatever you pass it it always returns `{}`, the test wont catch this change because it's true that in the test **and** in the repo.js script, it was called with `{}`, so what happens is that the test will still pass. It's a bit longer but much much safer to have it written out!

tests/unit/lambdas/repos.test.js
```diff
@@ -0,0 +1,117 @@
...
+    },
+    starCount: {
+      'en-US': 8724
+    }
+  }
+}
+
+let mockedContentfulRepos = [contentfulRepoOne, contentfulRepoTwo]
+
+describe('Repos lambda', () => {
```
omonk: Can we change this to `Github lambda - Repos util`, the repos script isn't a lambda, it just a dependency of one

tests/unit/lambdas/repos.test.js
```diff
@@ -0,0 +1,117 @@
...
+  let mockedEnvironment
+
+  beforeEach(() => {
+    mockedEnvironment = {
+      getEntries: jest.fn().mockReturnValue({ items: mockedContentfulRepos })
```
omonk: You can just pass `[contentfulRepoOne, contentfulRepoTwo]` here instead of defining 
```
let mockedContentfulRepos = [contentfulRepoOne, contentfulRepoTwo]
```

tests/unit/lambdas/repos.test.js
GBouffard: Yes it is. Good call. Thanks 👍

tests/unit/lambdas/repos.test.js
GBouffard: Sure. I wanted to separate with names for my own understanding but not a problem :)

tests/unit/lambdas/repos.test.js
GBouffard: True. It isn't. Doesn't return a status with body. Changed.

tests/unit/lambdas/repos.test.js
GBouffard: True. No need for a definition 👍

PR #351, opened 2019-08-21
update how we work section with new typography

omonk:
## Description - [Trello ticket](https://trello.com/c/vsJH7RoQ/132-revisit-a-few-more-updates-to-how-we-work-section-on-speciality-page)

Opening PR for deploy preview so James can take a look

### DIFF
--- a/src/components/Speciality/HowWeWorkWith.js
+++ b/src/components/Speciality/HowWeWorkWith.js
@@ -1,11 +1,12 @@
-import React, { Fragment } from 'react'
+import React from 'react'
 import styled from 'styled-components'
 import { Row, Col, Grid } from '../grid'
 import breakpoint from 'styled-components-breakpoint'
+import ReactMarkdown from 'react-markdown'
 
+import CustomisedBulletpoint from '../Common/CustomisedBulletpoint'
 import GreyBackground from '../Common/GreyBackground'
-import { SectionTitle, BodyPrimary } from '../Typography'
-import Hr from '../Common/Hr'
+import { SectionTitle, BodyPrimary, Subtitle } from '../Typography'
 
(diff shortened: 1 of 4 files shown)


src/templates/speciality/schemas.js
```diff
@@ -196,16 +196,9 @@ const howWeWorkSchema = {
     },
     howWeWorkWithCopy: {
       type: 'string'
-    },
-    howWeWorkWithPractises: {
-      type: 'array'
     }
```
GBouffard: Why did you remove `howWeWorkWithPractises`? It still should be in the query no?

src/components/Speciality/HowWeWorkWith.js
```diff
@@ -20,35 +23,73 @@ const StyledRow = styled(Row)`
...
 
 const HowWeWorkWithSection = ({
-  howWeWorkWithTitle,
-  howWeWorkWithCopy,
-  howWeWorkWithPractises
+  howWeWorkWithTitle: title,
+  howWeWorkWithCopy: copy,
+  howWeWorkWithPractises: practises = []
 }) => {
+  const hasPractises = practises && practises.length
```
GBouffard: I would move the `practises.length > 0` in this logic so you can use `hasPractises &&` below. It's nicer.

src/templates/speciality/schemas.js
omonk: I don't think we need it in here if its not required... 🤔 It wasn't working for me when I tried to just remove the practises in the `required` key

src/components/Speciality/HowWeWorkWith.js
omonk: @GBouffard nice spot, have updated

PR #352, opened 2019-08-21
Github lambda - Meta util test

GBouffard:
## Description - [Github lambda - Meta util test](https://trello.com/c/234HlV5X/635-github-lambda-meta-util-test)

Write the test

## Checklist

### DIFF
--- a/tests/unit/lambdas/meta.test.js
+++ b/tests/unit/lambdas/meta.test.js
@@ -1,0 +1,44 @@
+/* eslint-env jest */
+import MetaUtil from '../../../src/functions/oss/meta'
+
+import ossUtils from '../../../src/functions/oss/utils'
+ossUtils.updateEntry = jest.fn().mockImplementation(() => Promise.resolve(null))
+
+process.env.LAMBDA_ENV = 'production'
+
+const sameGithubMetaData = {
+  openSourceMetaPullRequestsCount: 3886,
+  openSourceMetaReposCount: 1001
+}
+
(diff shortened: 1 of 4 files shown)


omonk:
> Why do we have `return fieldsAreEqual ? contentfulMetaData : githubMetaData` in `./src/functions/oss/meta.js`?
> if it's equal it means that contentfulMetaData == githubMetaData so it should simply be
> `return githubMetaData` unless I'm missing something. Am I wrong in wanting to change that?

Errr yeah this isn't really a major thing. As you said, whatever happens you get the correct data, regardless of if you check for `fieldsAreEqual` or just return `githubMetaData`. 

The "reporting" we're using is fairly basic to be honest. It might be nicer to log the updates somewhere such as slack etc. but for the moment it's ok.

tests/unit/lambdas/meta.test.js
```diff
@@ -0,0 +1,68 @@
+/* eslint-env jest */
+import MetaUtil from '../../../src/functions/oss/meta'
+
+import ossUtils from '../../../src/functions/oss/utils'
+ossUtils.updateEntry = jest.fn().mockImplementation(() => Promise.resolve(null))
+
+process.env.LAMBDA_ENV = 'production'
```
omonk: Can you move this into the `beforeEach` then delete it in the `afterEach` or `afterAll`. My concern is there may be some side-effects to other tests if we leave it in here.

PR #354, opened 2019-08-22
CMS Speciality Logo Colours

GBouffard:
## Description - [617](https://trello.com/c/tfnmREtE/617-create-functionality-for-adding-speciality-logo-fill-colours-via-cms)

Logo colour was already added a non required field in Contentful `Speciality` but not always filled.

- [x] added all the colors of our live specialities in Contentful.
- [x] removed specialities colors from codebase
- [x] update logic to retrieve specialities colors from Contentful.
- [x] refactor it all so it's nicer and more explicit

## Checklist

### DIFF
--- a/src/components/Header/TopNav/index.js
+++ b/src/components/Header/TopNav/index.js
@@ -2,12 +2,14 @@ import React from 'react'
 import styled from 'styled-components'
 import breakpoint from 'styled-components-breakpoint'
 import remcalc from 'remcalc'
-
+import find from 'lodash.find'
 import { StaticQuery, graphql } from 'gatsby'
+
 import LogoLink from './LogoLink'
 import ServiceLink from './ServiceLink'
 import OuterAnchorItem from './OuterAnchorItem'
 import Dropdown from './Dropdown'
+import { logoColors } from '../navLinksHelper'
 
 const StyledTopNavContainer = styled.nav`
   display: flex;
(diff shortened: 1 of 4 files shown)


src/components/Header/TopNav/index.js
```diff
@@ -113,21 +118,30 @@ const TopNavBranding = ({ path, slug }) => (
         ? slug
         : getService({ slug, map: specialitiesToServicesMap })
 
+      const fillColorInitial = isSpecialityPage
+        ? getColor(specialities.nodes, slug) ||
```
AniaMakes: so getColor returns either a colour or... undefined?

GBouffard: No but close: the colour is not a required field in Contentful and if it's not there, we still have the field but it returns null; in that case the logic that follows will still default back to `white`. Maybe we would want to make that field required in Contentful if a color is published.

src/components/Header/TopNav/index.js
```diff
@@ -37,6 +39,8 @@ const TopNavList = styled.ul`
 
 const getSlugs = (arr = []) => arr.map(({ slug }) => slug).filter(i => i)
 
+const getColor = (arr, slug) => find(arr, { slug: slug }).logoColour
```
omonk: `find` will return undefined if it can't find anything, can you do: 
`(find(arr, { slug }) || {}).logoColor`

GBouffard: Fair point...it should always find something but let's just safe-guard. We never know 👍

PR #355, opened 2019-08-23
590 Add storybook addon

SemraSh:
## Ticket - [590](https://trello.com/c/GIZ9yCdD/590-add-storybook-addon-to-document-the-common-components)

Added the [storybook-addon-react-docgen](https://github.com/hipstersmoothie/storybook-addon-react-docgen/) to show the props in the stories.


.storybook/config.js
```diff
@@ -23,3 +23,4 @@ window.___navigate = pathname => {
 
 configure(loadStories, module)
 addDecorator(withA11y)
+addDecorator(withPropsTable)
```
omonk: I think you can add the same options here that you add throughout the stories, just as a second argument.

```
addDecorator(withPropsTable, {
  props: {
    tableComponent: TableComponent,
  },
});
```

This saves us having to write it out for every component. I'm hoping that if you want a different table component, as you have in `BlogListing.stories.js`, that the props in the story will override this. 

Can you check and see what happens. 

Also as we wrap the stories in the GlobalStyle and ThemeProvider it gets put in the props table, I think we can remove it if we pass these names as `propTablesExclude` 
image

SemraSh: I have changed the decorator to add the TableComponent in the beginning, thank you for pointing out. 

Regarding excluding components. I need to import the components that are excluded and GlobalStyleComponent is not imported properly from styled-components, it is not meant to be used directly so I couldn't import and exclude it. I have excluded ThemeProvider.

PR #359, opened 2019-08-28
645 - SEO Meta Data - Services

omonk:
## Description - [Trello ticket](https://trello.com/c/9ZInQlOM/645-seo-meta-data-services)

- Update service template and training page queries to reflect new seoMetaData field in CMS
- Update `Head` to handle new `seoMetaData` prop, this gives us the ability to transition gradually from old to new seo meta data structuring. See comment in Head for more details.
- Added new fragment for SeoMetaData fields
- Updated some gatsby plugins due to some upstream issues with the sharp package that was causing the build to fail: https://github.com/gatsbyjs/gatsby/issues/16957 

## Checklist


src/components/Common/Head.js
```diff
@@ -14,29 +14,72 @@ const TITLE = graphql`
   }
 `
 
-const Head = ({ page }) => {
+/**
+ * Contentful doesn't give us https: in the image urls so
+ * we have to prefix them here for twitter et al to pick
```
GBouffard: `et al` typo

omonk:
@GBouffard I agree with you on the comment in general, I don't usually write comments this long and I got a bit carried away tbh! It's more to highlight that this function shouldn't be around for very long...

It will be removed once all of the CMS data has been sorted out.

PR #370, opened 2019-09-03
Fetch medium data from CMS

fabiommmoreira:
## Investigate gatsby medium source plugin issues - [588](https://trello.com/c/ozDzAeA3/658-investigate-gatsby-medium-source-plugin-issues)

Create a script that fetch post data from medium and publishes it to Contentful.
Change queries that used to require medium content to consume Contentful blog data instead.

## Checklist


omonk:
Very nice! 🎉

I'm not sure we need to add the json file to the repo as it will change pretty soon...

This is a nice start for the initial migration blog start, I believe we can use import section of medium-json.js for the contentful upload when we start to feed from the RSS feed.

fabiommmoreira:
> Very nice! 🎉
> 
> I'm not sure we need to add the json file to the repo as it will change pretty soon...
> 
> This is a nice start for the initial migration blog start, I believe we can use import section of medium-json.js for the contentful upload when we start to feed from the RSS feed.

@omonk Right, I'll remove the `json-file.json` and add a note on `medium-json.js` about how to set things up before running the script 👍 

And yes, we can definitely salvage some of this work for the blog migration 😉

PR #382, opened 2019-09-19
669 contact us page

omonk:
## Description - [Trello ticket](https://trello.com/c/9I5mHKOd/669-contact-us-page)

- Remove old page
- Add new page!

### DIFF
--- a/src/components/ContactUs/Map.js
+++ b/src/components/ContactUs/Map.js
@@ -3,32 +3,41 @@ import { generate } from 'shortid'
 import GoogleMapReact from 'google-map-react'
 import styled from 'styled-components'
 import RatioContainer from '../Common/RatioContainer'
-import Marker from './Marker'
+import MapMarker from './Marker'
 import mapStyles from './mapStyles'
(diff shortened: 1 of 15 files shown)


src/components/ContactUs/Map.js
```diff
@@ -0,0 +1,43 @@
...
+import GoogleMapReact from 'google-map-react'
+import styled from 'styled-components'
+import RatioContainer from '../Common/RatioContainer'
+import MapMarker from './Marker'
+import mapStyles from './mapStyles'
+import getLatLngMidpoint from '../../utils/getLatLngMidpoint'
+
+const StyledRatioContainer = styled(RatioContainer)`
+  > * {
+    position: absolute !important;
```
fabiommmoreira: 👀

omonk: ikr, to be fair its for the google maps element in the DOM which has inline styling applied to it, I could put an inner and outer wrapper on it but seems a bit of overkill...

PR #386, opened 2019-09-25
674 - speciality blog background color

omonk:
## Description - [Trello ticket](https://trello.com/c/ROW7kGFV/674-speciality-pages-visual-discrepancies-audit)

Update query to include new `blogBackgroundColor` field, adds conditional wrapper around blog and events sections in speciality

### DIFF
--- a/src/components/Common/BlogListing.js
+++ b/src/components/Common/BlogListing.js
@@ -5,22 +5,9 @@ import { format } from 'date-fns'
 
 import { Grid } from '../grid'
 import TitleAndMediaList from '../Common/TitleAndMediaList'
-import GreyBackground from './GreyBackground'
+import BackgroundColorWrapper from '../Common/BackgroundColorWrapper'
 
-const Wrapper = ({ bgColor, children }) => {
-  switch (bgColor) {
-    case 'Grey':
-      return <GreyBackground>{children}</GreyBackground>
-    default:
-      return <>{children}</>
-  }
-}
-const BlogListing = ({
-  title,
-  description,
-  posts,
-  blogBackgroundColor = 'White'
-}) => {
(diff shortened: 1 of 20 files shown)


src/components/Common/BlogListing.js
```diff
@@ -5,8 +5,22 @@ import { format } from 'date-fns'
...
+      return <GreyBackground>{children}</GreyBackground>
+    default:
+      return <>{children}</>
+  }
+}
+const BlogListing = ({
+  title,
+  description,
+  posts,
+  blogBackgroundColor = 'White'
```
fabiommmoreira: should be lowercase for consistency

fabiommmoreira:
```suggestion
  blogBackgroundColor = 'white'
```

src/components/Common/BlogListing.js
```diff
@@ -15,18 +29,20 @@ const BlogListing = ({ title, description, posts }) => {
...
-          mediaItems={mediaItems}
-          CTALink="/blog/"
-          external
-          CTAText="Visit our blog"
-        />
-      </Padding>
-    </Grid>
+    <Wrapper bgColor={blogBackgroundColor}>
+      <Grid>
+        <Padding vertical={{ desktop: 5, smallPhone: 3.5 }}>
```
fabiommmoreira: Can we put the Padding component above the Grid? That way we could keep the Grid > Row > Col scheme

omonk: Yeah, this is annoying as the TitleAndMediaList component has `RowLayout` internally. I would opt for removing it altogether but not sure where to put the padding instead as I don't really feel it should be on the grid

PR #388, opened 2019-09-30
611 meetup lambda test

omonk:
## Description - [Trello ticket](https://trello.com/c/At97k5x8/611-meetup-lambda-test)

- Create meetup lambda tests for oauth and callback
- A few updates to meetup-callback and meetup-oauth


src/functions/meetup-callback.js
```diff
@@ -1,39 +1,24 @@
 /* eslint-disable no-console */
 const URLSearchParams = require('url').URLSearchParams
 const Got = require('got')
+const { every: Every } = require('lodash')
```
fabiommmoreira:
```suggestion
const Every = require('lodash.every')
```

before: 70K
after: 13K :)

PR #397, opened 2019-10-15
693 - landing page amends

omonk:
## Description - [Trello ticket](https://trello.com/c/BupfbsWx/693-post-launch-landing-page-amends)


src/components/Homepage/Intro.js
```diff
@@ -37,6 +37,8 @@ const IntroRow = styled(Row)`
 `
 
 const StyledCardTitle = styled(CardTitle)`
+  font-weight: 400;
```
jeysal: Looks like this is in `CardTitle` as `is('regular')`?

jeysal: Kinda weird that an override is called "regular" :see_no_evil:

src/components/Homepage/OurWork.js
```diff
@@ -96,7 +97,24 @@ const AnimatedLink = styled(Link)`
...
+
+  ${breakpoint('tablet')`
+    margin-top: ${({ theme }) => theme.space[5]};
+  `}
+`
+
+const CaseStudyImageWrapper = styled.div`
+  background-color: #${({ backgroundColor }) => backgroundColor};
+
+  > * {
```
jeysal: Why is this necessary?

src/components/Homepage/OurWork.js
```diff
@@ -109,22 +127,20 @@ const CaseStudy = ({ caseStudy }) => {
...
-        width={100}
-        height={130}
-        color={posterColor}
-        image={alternativePreviewImage}
-      >
+      <Card width={100} height={130} color={posterColor}>
         <CardHeader backgroundColor={posterColor}>
           <section>
-            <BodyPrimary reverse={reverseColor} muted>
+            <BodyPrimary noPaddingTop reverse={reverseColor} muted>
```
jeysal: Using this as boolean attributes causes React warnings :/
Do we have any plans to improve the modifier handling to keep it easy to use but also avoid warnings

stories/__snapshots__/AboutUs.stories.snapshot
```diff
@@ -1028,26 +1028,6 @@ exports[`Storyshots AboutUs Partners 1`] = `
 }
 
 .c6 {
-  -webkit-flex-wrap: wrap;
```
jeysal: Yeah, as we've talked about, I don't feel that most of the storyshots provide a lot of value in their current state :sweat_smile:

src/components/Homepage/OurWork.js
omonk: Ah yes, should have put a comment in here. this is because at some sizes the gap between the image and the header section was big enough (1 pixel) to see the grey background of the section it's on. This just prevents that

src/components/Homepage/OurWork.js
omonk: as per your slack comment we could probably refactor to remove the styled-is?

stories/__snapshots__/AboutUs.stories.snapshot
omonk: ditto

src/components/Homepage/Intro.js
omonk: oh boy

jeysal: So shall we use `regular` here instead of repeating the 400 constant?

src/components/Homepage/OurWork.js
jeysal: Can we put it on the child itself?

PR #398, opened 2019-10-21
preparations for 404 revamp

jeysal:
## Description

This mostly just refactors the `path` matching from `Branding` to React contexts provided by the pages and adds tests for the new logic (the old had none). This will allow the 404 page to also use the logo style of the homepage.
Made a few other small fixes / config improvements along the way.
Definitely review this commit by commit, otherwise the changes will seem completely random :sweat_smile:


.jest/filenameTransform.js
```diff
@@ -0,0 +1,7 @@
+const { basename } = require('path')
+
+module.exports = {
+  process(_src, filename) {
+    return 'module.exports = ' + JSON.stringify(basename(filename)) + ';'
```
jeysal: Unlike `fileMocks.js`, this allows us to distinguish between different asset files, i.e. test that the code loads the correct asset, not just any asset.

src/context/PageContext.js
```diff
@@ -0,0 +1,7 @@
+import { createContext } from 'react'
+
+export const HomePageContext = createContext(/* boolean */ false)
+export const ModalPageContext = createContext(/* boolean */ false)
+export const LogoStyleContext = createContext(
+  /* 'default' | 'white' */ 'default'
```
jeysal: I really want TS here ^^

src/pages/events.js
```diff
@@ -220,92 +222,96 @@ const EventPage = ({
...
-          eventStructuredData.map(data => (
-            <script key={generate()} type="application/ld+json">
-              {JSON.stringify(data)}
-            </script>
-          ))}
-      </Helmet>
-      <Head seoMetaData={seoMetaData} />
-
-      <StyledBlueBackground>
+    <LogoStyleContext.Provider value="white">
```
jeysal: No one can read a diff like this, I assure that I've only wrapped it in the `Provider` :joy:

src/utils/tests/test-utils.js
```diff
@@ -1,17 +0,0 @@
...
-import theme from '../theme'
-
-const AllTheProviders = ({ children }) => {
-  return <ThemeProvider theme={theme}>{children}</ThemeProvider>
-}
-
-const customRender = (ui, options) =>
-  render(ui, { wrapper: AllTheProviders, ...options })
-
-// re-export everything
```
jeysal: IMO an abstraction layer entirely hiding @testing-library/react behind a utils file is not worth it here vs using the library that many people know directly. Also not all tests will actually need this.

src/components/Header/DesktopNav/LogoLink.js
```diff
@@ -36,20 +39,20 @@ const LogoWrapper = ({ isHomePage, children }) =>
...
+  fillColorHover = 'grey',
+  textColor = 'black'
 }) => {
   const [fillColor, setFillColor] = useState(fillColorInitial)
+  const logoStyle = useContext(LogoStyleContext)
 
-  if (isHomePage || isEventsPage) {
+  if (logoStyle === 'white') {
     return (
-      <LogoWrapper isHomepage={isHomePage}>
```
jeysal: `scrollTo` was broken BTW because of prop name mismatch, that is also fixed now

src/context/PageContext.js
omonk: Yah this would be a much safer way of doing things...

omonk:
This looks good, thanks for the fresh take on the header logos 🎉

Whilst we're updating here do you think it is worth doing it with the speciality pages also?

src/pages/events.js
omonk: Aha I trust you!

jeysal:
> Whilst we're updating here do you think it is worth doing it with the speciality pages also?

The logic seems somewhat more complicated, but I can take a look at that next

jeysal:
Pushed it to yldio as well so that Circle runs https://circleci.com/gh/yldio/yld.io/tree/698-preparations
Won't be a problem next time now that I have access

jeysal:
I've pushed a Jest config rework https://github.com/yldio/yld.io/pull/398/commits/febcbf99720fe93da5d3dda9e65abbfe0b929ccc
May not be ideal, but unifies as much as possible given the current test types

PR #399, opened 2019-10-22
404 revamp

jeysal:
## 404 revamp - [698](https://trello.com/c/xR9OmjEt/698-404-revamp)

## Checklist

- [x] Updating or creating a new page? Consider any SEO requirements such as:
  SEO not relevant for 404 page


src/pages/404.js
```diff
@@ -7,28 +7,75 @@ import styled from 'styled-components'
...
+const CopyText = props => <BodyPrimary {...props} reverse="true" />
+
+const LinksTitle = styled(BodyPrimary).attrs({
+  reverse: 'true',
+  bold: 'true'
+})`
+  padding-top: ${({ theme }) => theme.space[3]};
+`
+
+const LinkParagraph = styled('p')`
```
omonk: We have all of the typography in styled components [here](https://github.com/yldio/yld.io/tree/master/src/components/Typography), have a look at storybook to get a feel of which one you need to use here.

As a rule of thumb on the project we pretty much always use predefined typography components

src/pages/404.js
```diff
@@ -7,28 +7,75 @@ import styled from 'styled-components'
...
+
+const TitleHeadline = props => (
+  <SectionTitle {...props} as="h1" reverse="true" />
+)
+
+const CopyText = props => <BodyPrimary {...props} reverse="true" />
+
+const LinksTitle = styled(BodyPrimary).attrs({
+  reverse: 'true',
+  bold: 'true'
```
omonk: Not saying this is wrong but the standard on this project is to pass in props at JSX level rather than in styled-component methods. Do you know of any differences in the outcomes?

src/pages/404.js
```diff
@@ -7,28 +7,75 @@ import styled from 'styled-components'
...
+  width: [1, 1, 1, 1, 0.5, 0.5]
+})`
+  ${breakpoint('smallTablet')`
+    padding-top: ${({ theme }) => theme.space[4]};
+    padding-bottom: ${({ theme }) => theme.space[4]};
+  `}
   ${breakpoint('tablet')`
-    margin: ${({ theme }) => theme.space[8]} 0
+    padding-top: unset;
+    padding-bottom: unset;
```
omonk: Using breakpoint you can set ranges for the media queries, 

`breakpoint('smallTablet', 'tablet')` - this will achieve what you want without having to create a new media query and use `unset` value

src/pages/404.js
```diff
@@ -7,28 +7,75 @@ import styled from 'styled-components'
...
+  padding-top: ${({ theme }) => theme.space[3]};
+`
+
+const LinkParagraph = styled('p')`
+  padding-top: ${({ theme }) => theme.space[2]};
+`
+
+const NotFoundPageLink = styled(Link)`
+  text-decoration: underline;
+`
```
omonk: Same as above, I believe we have `UnstyledLink` available for this instead of using the gatsby exporting component

src/pages/404.js
```diff
@@ -56,40 +102,56 @@ const NotFoundPage = () => (
...
+              <html lang="en" />
+            </Helmet>
+            <StyledBlueBackground>
+              <Grid>
+                <StyledRow>
+                  <TextCol>
+                    <TitleHeadline>{title}</TitleHeadline>
+                    <ReactMarkdown
+                      renderers={{
+                        paragraph: CopyText
```
omonk: Is it not easier to just put `(...props) => <BodyPrimary {...props} reverse="true" />` here instead?

omonk:
Something funky happening with CircleCi... 
Currently snapshots is still pending and nothing has failed but in reality everything passed apart from testcafe: 

image

Going to run it again

src/pages/404.js
jeysal: I kinda like having all the text containers here together, and it avoids the lint error that was previously ignored :)

src/pages/404.js
jeysal: Can't find `UnstyledLink`, and `StyledLink` has a lot of styling / `::after` I'd need to get rid of (used that at first)

src/pages/404.js
jeysal: I don't think there are any - [styled-components docs](https://www.styled-components.com/docs/api#attrs)
It just feels like they belong here because `reverse` and `bold` are very much a styling concern.

src/pages/404.js
jeysal: The only one I can find that would be about right is `BodyPrimary`, but I have to undo the padding and set reverse again (instead of relying on the color from the blue background). Pushed it but not sure if that's worth it?
It seems that `BodyPrimary` is more for paragraphs in proper articles, not sure if there's a more plain styled `<p>`?

jeysal:
> Currently github is reporting that snapshots is still pending and nothing has failed but in reality everything passed apart from testcafe:

I think it's confused because `snapshots` no longer exists, it's `storyshots` now. Dunno if this requires some manual intervention in repo settings or something to get rid of the old `snapshots` check?

src/pages/404.js
omonk: It's nested quite deeply but they have modifiers, `noPadding` , `noPaddingTop` etc. 
Then you can pass reverse as prop also. 

https://github.com/yldio/yld.io/blob/blog-add-mdx/src/components/Typography/modifiers.js#L4

src/pages/404.js
omonk: Do you mean that props should only be used for actions that concern behaviour etc.? 

I would agree in general although in the interest of consistency I'd prefer if we kept them defined on component level

omonk:
Ah I see... but wasn't this branch made after we got rid of it? Not sure how to get rid of it as everything defined in the yaml config should be good enough. Maybe a caching issue...

src/pages/404.js
omonk: Ok cool, lets see what design says, should be fine

src/pages/404.js
jeysal: Cool, using them now

src/pages/404.js
jeysal: > Do you mean that props should only be used for actions that concern behaviour etc.?

Not at all, just that I don't want to split the styling with the `styled-components` template string part at the top and the `modifiers` at the bottom.
At least not on one element, other elements do still have styling props at the bottom, I didn't go that far.

PR #414, opened 2019-11-08
combine test:unit and test:interaction

jeysal:
should probably check that the total number of test files has stayed the same :sweat_smile:


jeysal:
18 "unit" + 2 "interaction" from #412 = 20 "test" here, so looks like no tests lost

PR #425, opened 2019-11-13
request all (if more than 1000) blog posts from Contentful

jeysal:

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,8 @@
+{
+  "presets": ["babel-preset-gatsby"],
+  "env": {
+    "test": {
+      "plugins": ["require-context-hook"]
+    }
+  }
+}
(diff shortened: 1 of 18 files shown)


jeysal:
`result: []`

jeysal:
Cause for the difference in transpilation causing errors during module loading: https://github.com/netlify/netlify-lambda/issues/200
Changed it back to `.babelrc` for now

omonk:
I think this is default webpack behaviour no? https://github.com/netlify/netlify-lambda#webpack-configuration
I actually contributed to the netlify-lambda package to update the cli options to enable not using the babelrc flag. Previously it was breaking because the babelrc in the project uses `babel-preset-gatsby` which throws errors when used when compiling the lamdbas

jeysal:
Yeah I noticed the `--babelrc false` as well but I don't think it is working? At least the behavior changed when I renamed from `.babelrc.js` to `.babelrc` so ...

omonk:
Ah right yeah, it not picking up the .babelrc.js as you mentioned... we could probably make a PR to the package to update include .babelrc.js as well

PR #430, opened 2019-11-18
respect blog post publish flag

jeysal:
## respect blog post publish flag - [Trello ticket](https://trello.com/c/Y4zJcgXd/716-blog-post-launch-tasks)

This will help if an auto-published post has problems - we can take it offline by removing the `publish` flag in Contentful until the issue is fixed.

I've marked "Reduce bloat of your Lambdas" as `publish: false` so we can check out how it looks on the preview.
Only after we've set it to `true` again, we can merge this (that's why it's in "draft" state).


jeysal:
set the test post to `true` again

PR #447, opened 2019-12-05
add rel-preconnect links

jeysal:


jeysal:
Trying to optimize Analytics/GTM resources like this is too unstable (apparently even behaves differently on Netlify vs local prod builds) and of questionable real benefit, let's keep this at the okayish score it currently has

PR #452, opened 2020-03-30
stabilize lighthouse

jeysal:


jeysal:
https://circleci.com/gh/yldio/yld.io/13667?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link not really guaranteed that it's actually fixed, might just be lucky, but merging it is not gonna make it *more* flaky, so let's see

PR #461, opened 2020-05-22
One unified website logo

TdfGomes:
## Description - [Trello ticket](https://trello.com/c/9vJbf96j/756-website-logo-inconsistencies)

The navbar logo is now white square by default.

## Checklist


jeysal:
Hi and thanks, can we please avoid changing the default style to `'squared'` and only override it to `'squared'` for the pages that need to be fixed? Reasoning is that
1) There's lots of pages that use the default style without explicitly declaring it, say e.g. https://deploy-preview-461--yldio.netlify.app/privacy-policy/ , that are affected by this change
2) If we were to make this change (which I don't recommend since more page/page template files use the default), it would be weird to have a `'default'` style that is not a default, so we'd at least need to rename it

src/components/Header/DesktopNav/LogoLink.js
```diff
@@ -54,28 +52,15 @@ const LogoBehavior = ({ children, onMouseEnter, onMouseLeave }) => {
...
-          textColor={textColor}
-        />
-      ) : isHomePage || fillColorInitial !== colors.text ? (
-        // Do not fall back to animated logo if there is a different color,
-        // since we cannot recolor that. Also force non-animated on home page.
-        <YldLogo css={iconStyles} fillColor={fillColor} />
-      ) : (
-        <img css={iconStyles} src={animatedLogo} alt="YLD animated logo" />
-      )}
+      (
```
jeysal, line 62: Leftover parentheses. Hopefully these are what causes the logo being slightly misaligned too

src/context/PageContext.js
```diff
@@ -7,12 +7,10 @@ export const HomePageContext = createContext(/* boolean */ false);
 export const ModalPageContext = createContext(/* boolean */ false);
 
 export const logoStyleDefaults = {
-  type: 'default', // 'default' | 'squared'
-
   fillColorInitial: colors.text, // CSS color
   fillColorHover: undefined, // CSS color | undefined (default fillColorInitial)
 
-  textColor: colors.white, // CSS color (squared only)
+  textColor: colors.white,
```
jeysal, line 13:
```suggestion
  textColor: colors.white, // CSS color
```

src/components/Header/Branding.js
```diff
@@ -58,9 +58,8 @@ const TopNavBranding = ({ slug }) => (
       });
 
       let {
-        type,
         fillColorInitial = logoStyleDefaults.fillColorInitial,
-        fillColorHover = fillColorInitial,
+        fillColorHover,
```
jeysal, line 62: Setting defaults again here is so that you can do partial overrides using a context provider. If we do `fillColorHover = logoStyleDefaults.fillColorHover` here now, we can omit the `fillColorHover: colors.grey,` in the context providers, and it won't break if you introduce a context provider without a `fillColorHover` in the future :)
(This is because React does not merge object properties on context value like it does for state values)

PR #462, opened 2020-06-03
Add a 404 blog page for not found posts

TdfGomes:
## Description - [Trello ticket](https://trello.com/c/utImNeUO/758-blog-custom-404-page)

This PR create a `/blog` prefixed 404 page for not found blog post

## Checklist


gatsby-node.js
```diff
@@ -306,3 +306,15 @@ exports.onPostBuild = async ({ graphql }) => {
 
   await writeFile('./public/meta.json', JSON.stringify({ allJobIds }));
 };
+
+exports.onCreatePage = async ({
```
jeysal, line 310: This seems like a bit of a hack - if we make the blog 404 page a template instead of a page and include it in the build using `createPage` like we do many times above, I believe we can give it a proper `matchPath` straight away and don't have to hack around like this?

TdfGomes: I followed the documentation example for this kind of situation. You can see it (here)[https://www.gatsbyjs.org/docs/creating-prefixed-404-pages-for-different-languages/]

src/components/Common/Forms.js
```diff
@@ -41,6 +44,14 @@ export const Input = styled.input`
   ${is('noBoxShadow')`
     box-shadow:none;
   `};
+
+  ${is('serachBox')`
```
jeysal, line 48: typo: `serachBox`

gatsby-node.js
jeysal: Hmm, my point still stands - for this use case it seems much cleaner to just create the page once from a template, instead of hooking into its creation, where you then delete it and re-create it differently. But if you want to keep it this way, let's at least put the source of this snippet into a comment I'd say

src/components/Common/Forms.js
jeysal: Also in other places, even file names, make sure to catch them all ^^

src/components/Common/Forms.js
```diff
@@ -71,3 +82,102 @@ export const Fieldset = styled.section`
...
+    }
+  }
+  ${is('notFound')`
+    pointer-events:none;
+    color: ${props => props.theme.colors.grey};
+    font-weight: 500;
+  `};
+`;
+
+export const SearchBox = forwardRef(
```
jeysal, line 137: This component and its styles is so big and advanced that it should def get its own file, not together with the simple form components.

src/utils/replaceAccentChars.js
```diff
@@ -0,0 +1,10 @@
+export default function replaceAccentChars(str) {
```
jeysal, line 1: Why is this needed?

src/pages/404.js
```diff
@@ -101,7 +55,12 @@ const NotFoundPage = () => (
         linksTitle,
       } = content;
       return (
```
jeysal, line 57: It seems the previous comment about this is not addressed yet:

>There's still a lot of code duplication in the /404 and /blog/404 pages - can we extract the >whole structure of this page into the /components/404 file and just pass the things that are >specific to the /404 or /blog/404 as props?
>
>That also means that /components/404 won't have to export 10 small but highly specific >components that need to be properly assembled, but instead exports a well-defined 404 layout >component - seems cleaner to me (:

src/components/Common/Forms.js
```diff
@@ -71,3 +82,102 @@ export const Fieldset = styled.section`
...
+  }
+  to {
+    opacity: 1;
+  }
+`;
+
+export const SearchField = styled.section`
+  position: relative;
+  margin-bottom: ${remcalc(36)};
+`;
```
jeysal, line 98: Is the whole search field design signed off by James? If not should probably ping him to review the design

TdfGomes: This for now is just an idea for this page, but definitely I will talk with him about this.

src/components/Common/Forms.js
TdfGomes: I just wanted to keep all form UI in the same place. But I agree with you

src/utils/replaceAccentChars.js
TdfGomes: This will replace accent characters like _ã_ from the search term and also from the `title` and `authorName` then it will be easy for the user search the right article.

src/components/Common/Forms.js
TdfGomes: I was talking with Hannah and it happens that James is no longer work on the site, so any sing off is her responsibility.

src/components/Common/SearchBox.js
```diff
@@ -0,0 +1,108 @@
...
+    return (
+      <SearchField>
+        <Input
+          ref={ref}
+          type="search"
+          placeholder={placeholder}
+          aria-label="Search"
+          onChange={handleChange}
+          noBoxShadow
+          searchBox
```
fabiommmoreira, line 84: This renders inconsistently on different browsers. Some thing I've noticed:

1. Input padding and border-radius ignored on Safari, causing a split distance between the input and the result and outline to be off
Screenshot 2020-06-16 at 11 07 06

2. We're letting the browser decide the input clear button styling: a blue cross that appears when hovered on Chrome; a cross inside a grey circle always visible when the input has some value on Safari; Inexistent on Firefox. For consistency sake we should probably hide it for all browsers IMO
Screenshot 2020-06-16 at 11 11 28

This is probably caused by having _searchbox_ as type. I guess we need some style overrides on the browser defaults.

src/components/Common/SearchBox.js
```diff
@@ -0,0 +1,108 @@
...
+        />
+        {isListBoxOpen && (
+          <Results role="listbox">
+            {results.length > 0 ? (
+              results.map(({ title, slug }, idx) => (
+                <Result key={idx}>
+                  <Anchor to={slug}>
```
fabiommmoreira, line 91: We've the anchors only on the text blocks. Shouldn't they comprise the list item itself?

src/components/404/Ui.js
```diff
@@ -0,0 +1,63 @@
...
+import styled from 'styled-components';
+import remcalc from 'remcalc';
+import { Row, Col } from '../grid';
+import BlueBackground from '../Common/BlueBackground';
+import { SectionTitle, BodyPrimary } from '../Typography';
+
+// Layout
+
+export const StyledBlueBackground = styled(BlueBackground)`
+  margin-top: -${remcalc(36)};
```
fabiommmoreira, line 13: unnecessary `remcalc`, you can use `theme.space[3]`

src/components/Common/Forms.js
```diff
@@ -1,7 +1,10 @@
...
 import remcalc from 'remcalc';
-import styled from 'styled-components';
+import styled, { keyframes } from 'styled-components';
 import is from 'styled-is';
 
+import Anchor from './Anchor';
+
 export const Checkbox = styled.input`
   appearance: none;
   width: ${remcalc(24)};
```
fabiommmoreira, line 10: `theme.space[2]`

src/components/Common/Forms.js
```diff
@@ -71,3 +82,102 @@ export const Fieldset = styled.section`
     margin-bottom: ${remcalc(24)};
```
fabiommmoreira, line 82: `theme.space[2]`

src/components/Common/Forms.js
```diff
@@ -71,3 +82,102 @@ export const Fieldset = styled.section`
...
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+`;
+
+export const SearchField = styled.section`
+  position: relative;
+  margin-bottom: ${remcalc(36)};
```
fabiommmoreira, line 97: `theme.space[3]`

src/components/Common/Forms.js
```diff
@@ -71,3 +82,102 @@ export const Fieldset = styled.section`
...
+`;
+
+export const Results = styled.ul`
+  width: 100%;
+  max-height: ${remcalc(200)};
+  list-style-type: none;
+  background: ${props => props.theme.colors.greyBg};
+  position: absolute;
+  left: 0;
+  top: ${remcalc(56)};
```
fabiommmoreira, line 107: 56px is off from our spacing system... shouldn't we use 54px -> `theme.space[4]` ?

src/components/Common/Forms.js
```diff
@@ -71,3 +82,102 @@ export const Fieldset = styled.section`
...
+  z-index: 10;
+  overflow-x: auto;
+  overflow-y: scroll;
+  will-change: opacity;
+  animation: ${fadeIn} ${props => props.theme.animations.long} linear;
+`;
+
+export const Result = styled.li`
+  padding: ${remcalc(8)} ${remcalc(24)};
```
fabiommmoreira, line 116: `theme.spacing[1.5] theme.spacing[2]`

src/components/Common/Forms.js
```diff
@@ -71,3 +82,102 @@ export const Fieldset = styled.section`
...
+  font-size: ${remcalc(18)};
+  line-height: ${remcalc(24)};
```
fabiommmoreira, line 118: theme 😬

src/components/Common/SearchBox.js
```diff
@@ -0,0 +1,108 @@
...
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+`;
+
+export const SearchField = styled.section`
+  position: relative;
+  margin-bottom: ${remcalc(36)};
```
fabiommmoreira, line 20: `theme.space[4]`

src/components/Common/SearchBox.js
```diff
@@ -0,0 +1,108 @@
...
+`;
+
+export const Results = styled.ul`
+  width: 100%;
+  max-height: ${remcalc(200)};
+  list-style-type: none;
+  background: ${props => props.theme.colors.greyBg};
+  position: absolute;
+  left: 0;
+  top: ${remcalc(56)};
```
fabiommmoreira, line 30: 56px is off from our spacing system... shouldn't we use 54px -> theme.space[4] ?

src/components/Common/SearchBox.js
```diff
@@ -0,0 +1,108 @@
...
+  overflow-x: auto;
+  overflow-y: scroll;
+  will-change: opacity;
+  animation: ${fadeIn} ${props => props.theme.animations.long} linear;
+`;
+
+export const Result = styled.li`
+  padding: ${remcalc(8)} ${remcalc(24)};
+  font-size: ${remcalc(18)};
+  line-height: ${remcalc(24)};
```
fabiommmoreira, line 41: use theme values

src/components/Common/SearchBox.js
TdfGomes: will do it

src/components/Common/Forms.js
TdfGomes: will do it for space but will leave the font related properties, because we don't have it in the theme

src/components/Common/Forms.js
TdfGomes: will do it

src/components/Common/SearchBox.js
TdfGomes: will do it

src/components/Common/SearchBox.js
TdfGomes: will do it but will leave the text properties like that, because we don't have theme values for that

src/components/Common/SearchBox.js
fabiommmoreira: looking good now with type _text_ 👍

src/components/Common/SearchBox.js
fabiommmoreira: now filling the whole list item ;)

PR #467, opened 2020-06-26
fix inconsistent headers

TdfGomes:
## Description - [Trello ticket](https://trello.com/c/lDjsTxJa/763-headers-on-yldio)

fix inconsistent headers

## Checklist


src/templates/blog-post.js
```diff
@@ -69,7 +69,7 @@ const BlogPostTemplate = ({
 
   return (
     <MDXProvider components={{ ...components, ...shortCodes }}>
-      <Layout location={location}>
+      <Layout location={location} slug="Blog">
```
fabiommmoreira, line 72: slugs are usually lowercase to be URL friendly 🤔  but since you're using this "slug" in URL there won't be an issue, but just for semantic coherence maybe just rename the prop to "title"

Issue #492, opened 2021-02-11
error when installing sharp 0.22.1 with node version > 13


LauraBeatris:
There's an error when installing **sharp v0.22.1** (Currently being fixed using a **resolution** in `package.json`) with Node versions > 13

I'm currently using Node.js version 14.15 and that's the error that I get when trying to install `sharp`:

![image](https://user-images.githubusercontent.com/48022589/107679389-44718680-6c7b-11eb-9e10-97abe22a7e14.png)

According to this issue https://github.com/lovell/sharp/issues/2005#issuecomment-564491092, support for **Node.js 13** was added in sharp **v0.23.2**, although our resolution is setting sharp to **v0.22.1** which doesn't support Node 13 >

LauraBeatris:
The error is happening not just in one dependency but in all of those who are **sharp < v0.23.2**

For instance, `favicons` seems to be a peer dependency from `gatsby-plugin-favicon` and it uses sharp **v0.22.1**

![image](https://user-images.githubusercontent.com/48022589/107680890-2f95f280-6c7d-11eb-8f11-9e17a49511c9.png)

Should we update sharp resolution to the last version? In that way, it will avoid the error when installing all packages that are using sharp

PR #493, opened 2021-02-11
build: set node version to 14

LauraBeatris:
## Description 

Set Node.js version to 14 (Last version) and [lock](https://tecadmin.net/lock-application-with-specific-node-version/) application to Node.js version 14.

It was being set to 10 and I believe that it's because that AWS Lambda didn't support Node.js > 12 at the time, refer to this [PR](https://github.com/yldio/yld.io/pull/413), but now AWS supports Node.js 14 https://aws.amazon.com/about-aws/whats-new/2021/02/aws-lambda-now-supports-node-js-14/

## How to test 
- Run `nvm use` and `nvm install` in case that you don't have this Node.js version
- Install dependencies & run build


LauraBeatris:
By updating Node.js version, it's also going to be necessary to update sharp version, refer to #492

package.json
```diff
@@ -182,7 +182,7 @@
     "webpack-node-externals": "^1.7.2"
   },
   "resolutions": {
-    "sharp": "0.22.1",
+    "sharp": "0.27.0",
```
LauraBeatris, line 185: By updating `Node.js` version, it's also going to be necessary to update `sharp` version, refer to #492

LauraBeatris:
There's also another env var defining the Node.js version for AWS lambda 

image

LauraBeatris:
I'm close this for now and create a branch in this repo with the same changes in order to run the CircleCI tests

PR #494, opened 2021-02-12
build: set node version to 14

LauraBeatris:
## Description 

Set Node.js version to 14 (Last version) and [lock](https://tecadmin.net/lock-application-with-specific-node-version/) application to Node.js version 14.

It was being set to 10 and I believe that it's because that AWS Lambda didn't support Node.js > 12 at the time, refer to this [PR](https://github.com/yldio/yld.io/pull/413), but now AWS supports Node.js 14 https://aws.amazon.com/about-aws/whats-new/2021/02/aws-lambda-now-supports-node-js-14/

## How to test 
- Run `nvm use` and `nvm install` in case that you don't have this Node.js version
- Install dependencies & run build


LauraBeatris:
There's also another env var defining the Node.js version for AWS Lambda 

image

LauraBeatris:
Before merging this PR, I'll like to discuss if we should keep `.nvmrc` to lock the Node.js version or centralize this in an environment variable called `NODE_VERSION` as suggested by @ricardofbarros 

One thing that I like about having a `.nvmrc` file is to be possible to easily set and lock a default Node.js version in the project by using the command `nvm use`

Here you can see what Netlify uses to define the Node.js version https://github.com/netlify/build-image/blob/xenial/included_software.md

ricardofbarros:
I'm fine with either approach @LauraBeatris, I didn't realize we already had a `.nvmrc` file.

So one argument is that if we move it to an env var we would have it centralized, which is nice. But as you said, we would lose some clarity while doing local development as it is easier to have a `.nvmrc` to keep track of which node version should the website run.

ricardofbarros (approved):
LGTM 🚀

sergioramos:
this is fantastic @LauraBeatris thank you 👍

ricardofbarros:
Btw good work @LauraBeatris  on having the sharp dependency sorted out. I've encountered this issue in the past, but I just brushed it aside.

LauraBeatris:
Thanks @ricardofbarros and @sergioramos 

Now that it's merged. I've updated the environment variable from Netfily that defines the Node.js version for the serverless functions, and then I'm going to re-run the pipelines.

